PR #6, opened 2017-01-20
Added two samples:

singhgag:
- Feature collection layer
- Feature collection layer (query)


dg0yal:
Same as iOS, copyright headers need to be updated. Everything else looks good and can be merged

PR #38, opened 2017-12-02
New Sample: WMS layer (URL)

gdhakal:
This sample demonstrates how to create a WMS layer using the service URL and the layer names that are available in the service.


gdhakal:
@dg0yal The wms layer draws on map view successfully without the ATS exception, which is very interesting to me as it did not work in iOS without the ATS exception.

dg0yal:
Doesn't work for me. What version of the macOS operating system and mac SDK are you using?
My machine is on 10.12.6 and the SDK is 10.13

PR #47, opened 2017-12-08
New Sample: Symbolize Shapefile

nixta:


dg0yal:
Can you update this PR with just the data files we need for the sample

nixta:
@dg0yal Shapefiles removed.

PR #66, opened 2018-02-19
Simplification suggestions

nixta:
Hey @gdhakal.

This PR has some suggestions from playing with the sample.

I felt we could do better with setting the colors. While the code you had worked, it isn't very safe. Also, this way it's easier to get a color with an alpha component.

I also noticed the sliders only updated when you released them (unlike the iOS app). There's a `continuous` setting on the slider control in the storyboard that changes this.

I also tidied up the controls stack view a little bit.

Let me know what you think.


gdhakal:
Looks good 👍 Thanks @nixta

PR #69, opened 2018-02-21
Read the AGSViewshed colors and reflect them in the UI

nixta:


nixta:
@gdhakal Could you take a look at this and confirm it works as expected for you please? If you change the colors, exit the sample, then come back to the sample, the UI should refer to the colors that were set previously (they're global on AGSViewshed so persist between viewshed samples).

Thanks!

PR #77, opened 2018-05-31
New Sample: Open a scene (portal item)

philium:
This sample demonstrates how to open a web scene from a portal item.


arcgis-runtime-samples-macos/Scenes/Open a scene (portal item)/OpenSceneViewController.swift
```diff
@@ -0,0 +1,36 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import Cocoa
+import ArcGIS
+
+class OpenSceneViewController: NSViewController {
+    var sceneView: AGSSceneView {
```
nixta: We don't want to promote setting the top-level view as an `AGSSceneView` (or `AGSMapView`) since we don't want to promote using an `AGSGeoView` as a container view. Very rarely will a customer have nothing but a map or scene view in a view hierarchy.

Better to maintain the default top-level view as it is, and add an `AGSSceneView` sub-view that fills it, referencing it as an `IBOutlet`. See [this macOS](https://github.com/Esri/arcgis-runtime-samples-macos/blob/v.next/arcgis-runtime-samples-macos/Scenes/Display%20a%20scene/DisplaySceneViewController.swift#L22) and [this iOS](https://github.com/Esri/arcgis-runtime-samples-ios/blob/35f06faa3c5a357b3690e9ef7d9f77a21f42b4e2/arcgis-ios-sdk-samples/Scenes/Open%20a%20scene%20(portal%20item)/OpenSceneViewController.swift#L20) sample.

PR #80, opened 2018-06-01
Updates copyright info

philium:
The copyright info has been updated to include the current year and the Apache License notice has been removed.

Before | After
------ | -----
about before | about after
info before | info after


philium:
Thank you @dg0yal! I'll make the changes you requested.

I did notice that "All rights reserved." is used on the [Esri Legal](https://www.esri.com/en-us/legal/copyright-trademarks/) webpage.

PR #81, opened 2018-06-06
Converts custom colors from methods to properties

philium:
This matches Apple's convention of using class properties for custom colors.


dg0yal:
Good change 👍

PR #83, opened 2018-06-08
Adds a Distance Measurement Analysis sample

philium:
This sample demonstrates measuring 3D distances between two points in a scene.


arcgis-runtime-samples-macos/Analysis/Distance measurement analysis/DistanceMeasurementAnalysisViewController.swift
```diff
@@ -0,0 +1,100 @@
...
+    let scene = AGSScene(basemap: .imagery())
+    var locationDistanceMeasurement = AGSLocationDistanceMeasurement(startLocation: defaultStartPoint, endLocation: defaultEndPoint)
+    
+    var distanceWindowController: NSWindowController?
+    
+    override init(nibName nibNameOrNil: NSNib.Name?, bundle nibBundleOrNil: Bundle?) {
+        fatalError("init(nibName:bundle:) has not been implemented")
+    }
+    
+    required init?(coder: NSCoder) {
```
nixta: Could this not be done in `viewDidLoad`? I believe other samples do that. This way we can avoid overriding any init() methods on `NSViewController`.

arcgis-runtime-samples-macos/Analysis/Distance measurement analysis/DistanceMeasurementAnalysisViewController.swift
```diff
@@ -0,0 +1,100 @@
...
+                AGSArcGISTiledElevationSource(url: URL(string: "https://tiles.arcgis.com/tiles/d3voDfTFbHOCRwVR/arcgis/rest/services/MNT_IDF/ImageServer")!)
+            ]
+        )
+        let buildingsLayer = AGSArcGISSceneLayer(url: .brestBuildingsService)
+        
+        scene.baseSurface = surface
+        scene.operationalLayers.add(buildingsLayer)
+    }
+    
+    override func viewDidAppear() {
```
nixta: I like that you've broken out the `AGSGeoViewTouchDelegate` code into a separate extension. But (if we continue with the `NSPanel`) can we move these panel control methods into another extension which comes at the end of the file? As they are, these functions distract from the Runtime code in the `ASGGeoViewTouchDelegate` extension, which is (along with the init() code above) what we want people to see.

arcgis-runtime-samples-macos.xcodeproj/project.pbxproj
```diff
@@ -333,9 +333,19 @@
...
+		4C2244AF20C9B0F6005E6DEE /* URL.swift in Sources */ = {isa = PBXBuildFile; fileRef = 4C2244AE20C9B0F6005E6DEE /* URL.swift */; };
+		4C2244B320C9C39C005E6DEE /* AGSSurface.swift in Sources */ = {isa = PBXBuildFile; fileRef = 4C2244B220C9C39C005E6DEE /* AGSSurface.swift */; };
+		4C2244B520C9C3D7005E6DEE /* AGSAnalysisOverlay.swift in Sources */ = {isa = PBXBuildFile; fileRef = 4C2244B420C9C3D7005E6DEE /* AGSAnalysisOverlay.swift */; };
+		4C2244B720C9CB57005E6DEE /* DistanceViewController.swift in Sources */ = {isa = PBXBuildFile; fileRef = 4C2244B620C9CB57005E6DEE /* DistanceViewController.swift */; };
+		4C2244B920C9DF93005E6DEE /* DistancePanel.swift in Sources */ = {isa = PBXBuildFile; fileRef = 4C2244B820C9DF93005E6DEE /* DistancePanel.swift */; };
+		4C2244BA20CB0299005E6DEE /* DistanceViewController.swift in CopyFiles */ = {isa = PBXBuildFile; fileRef = 4C2244B620C9CB57005E6DEE /* DistanceViewController.swift */; };
+		4C2244BC20CB1811005E6DEE /* README.md in Resources */ = {isa = PBXBuildFile; fileRef = 4C2244BB20CB1811005E6DEE /* README.md */; };
```
nixta: This should not be included in the project.

If we do want to include them in the project (which I _do_ find useful while working with Xcode), it should not have any active target memberships.

Please open an issue proposing that we bring all READMEs into the iOS and macOS sample projects with no target to let others chime in on this.

arcgis-runtime-samples-macos/Analysis/Distance measurement analysis/DistanceMeasurementAnalysisViewController.swift
```diff
@@ -0,0 +1,100 @@
...
+import Cocoa
+import ArcGIS
+
+private let defaultStartPoint = AGSPoint(x: -4.494677, y: 48.384472, z: 24.772694, spatialReference: .wgs84())
+private let defaultEndPoint = AGSPoint(x: -4.495646, y: 48.384377, z: 58.501115, spatialReference: .wgs84())
+
+class DistanceMeasurementAnalysisViewController: NSViewController {
+    @IBOutlet weak var sceneView: AGSSceneView! {
+        didSet {
+            sceneView.touchDelegate = self
```
nixta: I generally really like this pattern, but it's a bit non-linear when teaching the Runtime. It's usually a bit clearer to group this stuff with the rest of the stuff currently in `init(coder)`. Could we do that?

Take a look at the [Line of Sight](https://github.com/Esri/arcgis-runtime-samples-macos/blob/v.next/arcgis-runtime-samples-macos/Analysis/Line%20of%20sight%20(location)/LineOfSightLocationViewController.swift) sample.

In there, we _do_ break out the lineOfSight analysis, but we're initializing the line of sight later as part of the `AGSGeoViewTouchDelegate` methods (more on that in a another review comment).

arcgis-runtime-samples-macos/Analysis/Distance measurement analysis/DistanceMeasurementAnalysisViewController.swift
```diff
@@ -0,0 +1,100 @@
...
+            sceneView.scene = scene
+            sceneView.setViewpointCamera(AGSCamera(lookAt: defaultStartPoint, distance: 200, heading: 0, pitch: 45, roll: 0))
+            sceneView.analysisOverlays.add(AGSAnalysisOverlay(analyses: [locationDistanceMeasurement]))
+        }
+    }
+    
+    let scene = AGSScene(basemap: .imagery())
```
nixta: Less critical than with the sceneView setup, but see above - it might be a better teaching resource if this initialization were moved in-line with the overall setup of both scene and sceneView.

arcgis-runtime-samples-macos/Analysis/Distance measurement analysis/DistanceMeasurementAnalysisViewController.swift
```diff
@@ -0,0 +1,100 @@
...
+            let distanceViewController = windowController.contentViewController as? DistanceViewController else {
+                preconditionFailure()
+        }
+        distanceViewController.locationDistanceMeasurement = locationDistanceMeasurement
+        distanceWindowController = windowController
+    }
+}
+
+extension DistanceMeasurementAnalysisViewController: AGSGeoViewTouchDelegate {
+    func geoView(_ geoView: AGSGeoView, didTouchDownAtScreenPoint screenPoint: CGPoint, mapPoint: AGSPoint, completion: @escaping (Bool) -> Void) {
```
nixta: While testing the sample, if I have a measurement in progress and single-click, the measurement graphic remains on screen but the panel updates with null measurements (is this a bug with the measurement tool?).

But… scene interactions are unexpected (dragging around the scene does not pan it).

I wonder if a different UX flow, again like in the [Line of Sight](https://github.com/Esri/arcgis-runtime-samples-macos/blob/5d8f21fe76bc8635137cd50b162db78c661f5fe5/arcgis-runtime-samples-macos/Analysis/Line%20of%20sight%20(location)/LineOfSightLocationViewController.swift#L77-L92) sample, might be appropriate. There a click places one end of the measurement and just moving the mouse moves the other end. This felt more natural on a desktop interface, and had the side-effect of freeing the drag operation to actually move the scene around.

That might avoid the issue I'm seeing with null measurements displayed while a measurement graphic is displayed.

arcgis-runtime-samples-macos/Extensions/ArcGIS/AGSAnalysisOverlay.swift
```diff
@@ -0,0 +1,27 @@
...
+// limitations under the License.
+//
+
+import ArcGIS
+
+extension AGSAnalysisOverlay {
+    /// Creates an analysis overlay with the given analyses.
+    ///
+    /// - Parameter analyses: A sequence of `AGSAnalysis` objects.
+    convenience init<S: Sequence>(analyses: S) where S.Element == AGSAnalysis {
```
nixta: I think this is a good convenience initializer for an actual app (and something we might see in the example apps), but for sample apps we don't want to hide the vanilla SDK workflows.

I think we should remove this, but open to discussion with the team.

arcgis-runtime-samples-macos/Extensions/ArcGIS/AGSSurface.swift
```diff
@@ -0,0 +1,27 @@
...
+// limitations under the License.
+//
+
+import ArcGIS
+
+extension AGSSurface {
+    /// Creates a surface with the given elevation sources.
+    ///
+    /// - Parameter elevationSources: A sequence of `AGSElevationSource` objects.
+    convenience init<S: Sequence>(elevationSources: S) where S.Element == AGSElevationSource {
```
nixta: As above: I think this is a good convenience initializer for an actual app (and something we might see in the example apps), but for sample apps we don't want to hide the vanilla SDK workflows.

I think we should remove this, but open to discussion with the team.

arcgis-runtime-samples-macos/Analysis/Distance measurement analysis/DistanceMeasurementAnalysisViewController.swift
```diff
@@ -0,0 +1,100 @@
...
+        fatalError("init(nibName:bundle:) has not been implemented")
+    }
+    
+    required init?(coder: NSCoder) {
+        super.init(coder: coder)
+        
+        let surface = AGSSurface(
+            elevationSources: [
+                AGSArcGISTiledElevationSource(url: .terrain3DService),
+                AGSArcGISTiledElevationSource(url: URL(string: "https://tiles.arcgis.com/tiles/d3voDfTFbHOCRwVR/arcgis/rest/services/MNT_IDF/ImageServer")!)
```
nixta: I like the idea of providing a central collection of sample service URLs, but we now have both a convenience property and a full URL in adjacent lines which gets murky. A developer looking at this will ask "What's special about one but not the other??".

Instead, could we put the URLs inline here for all 3 URLs, and open an issue to discuss refactoring these sample URLs across the whole sample app please (e.g. is an extension on `URL` the right place to put it?)?

arcgis-runtime-samples-macos/Analysis/Distance measurement analysis/DistancePanel.swift
```diff
@@ -0,0 +1,23 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+import AppKit
+
+class DistancePanel: NSPanel {
```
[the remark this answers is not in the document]
philium: The problem I see with an overlay is twofold:

1. It covers a portion of the graph. In the case of the Viewshed (location) sample, a large portion.
1. It isn't movable.

Were you seeing an issue with the distance panel appearing on a different display than the app's main window, or have you seen that happen when using panels in the past? In my testing, the distance panel always displayed on the same screen as the sample.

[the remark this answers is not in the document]
nixta: I saw this with the app. Either on the first or second run the app opened on my secondary display but the panel appeared on my main display.

After some more tinkering, this appears to be dependent on whether the app is Quit properly (then the panel seems to appear on the right display) or if you stop the debugger in Xcode while the app is on display 2 (on next run, the app will start on display 2 but the panel will appear on display 1).

arcgis-runtime-samples-macos/Analysis/Distance measurement analysis/DistanceMeasurementAnalysisViewController.swift
philium: > You can override this method to perform tasks to immediately follow the setting of the view property.
Typically, your override would perform one-time instantiation and initialization of the contents of the view controller’s view.

Though a common practice, `viewDidLoad()` is not meant to be used as an initializer for non-view properties of a view controller.

This sample doesn't demonstrate it, but there are advantages to creating the model before the view loads. The initializer is called earlier than `viewDidLoad()`. In the case of a remote resource that needs to be loaded, requesting it to be loaded in the initializer increases the likelihood that the resource will be loaded before the view displays.

nixta: Understood, but we end up fragmenting the Runtime sample code because you can't put `AGSSceneView` code in here, so we have some code in `init()`, some in the getter for `sceneView`, and some in the variable declarations.

arcgis-runtime-samples-macos/Analysis/Distance measurement analysis/DistanceMeasurementAnalysisViewController.swift
nixta: A local constant, you mean, as opposed to an extension to `URL`?

nixta:
Great stuff. Thanks for all the work on this.

PR #87, opened 2018-06-14
Uses previously defined static URLs in all samples

philium:
This also removes the App Transport Security exception for arcgis.com as all arcgis.com URLs now use https.

### DIFF
--- a/arcgis-runtime-samples-macos/Info.plist
+++ b/arcgis-runtime-samples-macos/Info.plist
@@ -43,15 +43,6 @@
 				<key>NSThirdPartyExceptionRequiresForwardSecrecy</key>
 				<false/>
 			</dict>
-			<key>arcgis.com</key>
-			<dict>
-				<key>NSIncludesSubdomains</key>
-				<true/>
-				<key>NSThirdPartyExceptionAllowsInsecureHTTPLoads</key>
-				<true/>
-				<key>NSThirdPartyExceptionRequiresForwardSecrecy</key>
-				<false/>
-			</dict>
 			<key>arcgisonline.com</key>
 			<dict>
 				<key>NSIncludesSubdomains</key>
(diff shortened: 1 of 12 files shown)


rolson:
> This also removes the App Transport Security exception for arcgis.com as all arcgis.com URLs now use https.

Sometimes a map will list urls to layers and those will not have https.

arcgis-runtime-samples-macos/Info.plist
```diff
@@ -43,7 +43,7 @@
 				<key>NSThirdPartyExceptionRequiresForwardSecrecy</key>
 				<false/>
 			</dict>
-			<key>arcgis.com</key>
```
nixta: We should not make this change. As @rolson says, one could load a Web Map definition (over HTTPS) which itself includes references to hosted services using just HTTP.

rolson: @dg0yal and @njarecha can you confirm my and Nick's suspicion here?

philium: @rolson @dg0yal @njarecha Have you had a chance to look into this? It would be nice to remove this exception if it is no longer needed.

dg0yal: We need to keep the `NSThirdPartyExceptionAllowsInsecureHTTPLoads` for reasons mentioned above. But we can remove the `NSThirdPartyExceptionRequiresForwardSecrecy` exception because arcgis.com and arcgisoline.com support the latest ciphers (they didn't a few years ago).

philium: @dg0yal Thank you. I'll update the PR.

nixta: @philium Based off @dg0yal's reply, we can get rid of `NSThirdPartyExceptionRequiresForwardSecrecy` for `arcgisonline.com` as well as `arcgis.com`.

philium: @nixta My thought was to do that in a separate PR. Do you want me to do it in this one?

nixta: Oh. I see. None of the URLs modified here are `arcgisonline.com`. Ok, separate PR sounds good.

PR #88, opened 2018-06-14
Prevents "Source File:" label from being truncated

philium:
This would happen when source files names were on the longer side.

Before:
screen shot 2018-06-14 at 1 31 36 pm

After:
screen shot 2018-06-14 at 1 32 19 pm


esreli:
does this PR reference an issue?

esreli:
https://devtopia.esri.com/runtime/common-samples/issues/557#issuecomment-1213802

philium:
@nixta I was going for the simplest fix possible. The user can still see the full filename in the pop up menu.

We could certainly allow the pop up button to expand as needed, but that would be a larger change.

philium:
@nixta Could you please approve this?

PR #89, opened 2018-06-14
Ensure source label always displayed and filename displayed as possible

nixta:
@philium: See if you like this. No worries if not.


arcgis-runtime-samples-macos/Base.lproj/Main.storyboard
```diff
@@ -1081,10 +1082,21 @@
                                 </webPreferences>
                             </webView>
                             <customView translatesAutoresizingMaskIntoConstraints="NO" id="OR7-Wf-OcB">
```
philium: Could you please get rid of this view? The stack view has taken it's place.

nixta:
@philium Sure thing. PR updated.

PR #99, opened 2018-06-19
Makes MapRotationViewController update UI on main thread

philium:


rolson:
looks fine - just wondering for sample code if breaking it out into a separate function is helpful vs harmful for simplicity of a sample

rolson:
it is more readable that way, yes

PR #106, opened 2018-06-22
Adds the Query a Map Image Sublayer sample

philium:


arcgis-runtime-samples-macos/Layers/Query a map image sublayer/QueryMapImageSublayerViewController.swift
```diff
@@ -0,0 +1,196 @@
...
+        
+        // Add the graphics overlay to the map view.
+        mapView.graphicsOverlays.add(graphicsOverlay)
+        
+        enableControlsIfNeeded()
+    }
+    
+    /// Called in response to the layer loading successfully.
+    func layerDidLoad() {
+        imageLayer.mapImageSublayers.forEach { ($0 as? AGSArcGISMapImageSublayer)?.load() }
```
nixta: You probably want to use `AGSLoadObjects()` here.

Perhaps something like:

``` Swift
var subLayers:[AGSArcGISMapImageSublayer] {
    return imageLayer.mapImageSublayers as? [AGSArcGISMapImageSublayer] ?? []
}

func layerDidLoad() {
    AGSLoadObjects(subLayers) { [weak self] _ in
        self?.enableControlsIfNeeded()
    }
}
```
and
``` Swift
func enableControlsIfNeeded() {
    guard isViewLoaded, imageLayer.loadStatus == .loaded, subLayers.filter({ subLayer in
        return subLayer.loadStatus == .loaded
    }).count > 0 else { return }

    populationTextField.isEnabled = true
    populationTextField.window?.makeFirstResponder(populationTextField)
    queryButton.isEnabled = true
}
```

arcgis-runtime-samples-macos/Layers/Query a map image sublayer/QueryMapImageSublayerViewController.swift
```diff
@@ -0,0 +1,196 @@
...
+            alert.messageText = "The population value must be numeric."
+            alert.addButton(withTitle: "OK")
+            alert.beginSheetModal(for: self.view.window!)
+        }
+    }
+    
+    /// The population value provided by the user.
+    var populationValue: Int? {
+        didSet {
+            populationValueDidChange()
```
nixta: This is only called from here - should we just inline the function's contents?

arcgis-runtime-samples-macos/Layers/Query a map image sublayer/QueryMapImageSublayerViewController.swift
```diff
@@ -0,0 +1,196 @@
...
+    
+    /// Called when a feature query of a feature table finishes successfully.
+    ///
+    /// - Parameters:
+    ///   - featureTable: The feature table that was queried.
+    ///   - result: The feature query result.
+    func featureTable(_ featureTable: AGSFeatureTable, featureQueryDidSucceedWith result: AGSFeatureQueryResult) {
+        let symbol = makeSymbol(featureTable: featureTable)
+        let graphics: [AGSGraphic] = result.featureEnumerator().map {
+            AGSGraphic(geometry: ($0 as! AGSFeature).geometry, symbol: symbol, attributes: nil)
```
nixta: We can avoid the `!` unwrap (and infer type) with:

``` Swift
let graphics = result.featureEnumerator().allObjects.map {
    AGSGraphic(geometry: $0.geometry, symbol: symbol, attributes: nil)
}
```

arcgis-runtime-samples-macos/Layers/Query a map image sublayer/README.md
```diff
@@ -0,0 +1,43 @@
...
+The sublayers (represented by `AGSArcGISMapImageSublayer`), expose a `AGSServiceFeatureTable` through a `table` property. A query is created to find features with a population greater than the value provided that
+are also in the current map extent. Since the `cities`, `counties`, and `states` tables all have a `POP2000` field, they can all execute the same query.
+The features selected by the query are selected in the corresponding layers to display the results in the map.
+
+
+## Relevant API
+
+ - `AGSServiceFeatureTable`
+ - `AGSArcGISMapImageLayer`
+ - `AGSArcGISMapImageSublayer.load(completion:)`
```
nixta: If you use `AGSLoadObjects()` as suggested elsewhere, this is no longer relevant.

arcgis-runtime-samples-macos/Layers/Query a map image sublayer/QueryMapImageSublayerViewController.swift
```diff
@@ -0,0 +1,196 @@
...
+        
+        let populationQuery = AGSQueryParameters()
+        populationQuery.whereClause = "POP2000 > \(populationValue)"
+        populationQuery.geometry = mapView.currentViewpoint(with: .boundingGeometry)?.targetGeometry
+        
+        for table in [citiesTable, statesTable, countiesTable] {
+            table?.queryFeatures(with: populationQuery) { [weak self, weak table] (result, error) in
+                guard let featureTable = table else { return }
+                if let result = result {
+                    self?.featureTable(featureTable, featureQueryDidSucceedWith: result)
```
nixta: Should we inline this and the `else` function below since they're not used elsewhere. I think it'll make it more readable. We can avoid the optional symbol with:

``` Swift
for table in [citiesTable, statesTable, countiesTable] {
    table?.queryFeatures(with: populationQuery) { [weak self, weak table] (result, error) in
        guard let featureTable = table, let symbol = self?.makeSymbol(featureTable: featureTable) else { return }
        if let result = result {
            let graphics: [AGSGraphic] = result.featureEnumerator().map {
                AGSGraphic(geometry: ($0 as! AGSFeature).geometry, symbol: symbol, attributes: nil)
            }
            self?.graphicsOverlay.graphics.addObjects(from: graphics)
        } else if let error = error {
            print("\(featureTable.tableName) feature query failed: \(error)")
        }
    }
}
```

See also other review note on avoiding the `as! AGSFeature` also.

arcgis-runtime-samples-macos/Layers/Query a map image sublayer/QueryMapImageSublayerViewController.swift
```diff
@@ -0,0 +1,196 @@
...
+            } else {
+                self?.layerDidLoad()
+            }
+        }
+    }
+    
+    /// The map view managed by the view controller.
+    @IBOutlet weak var mapView: AGSMapView!
+    @IBOutlet weak var populationTextField: NSTextField!
+    @IBOutlet weak var queryButton: NSButton!
```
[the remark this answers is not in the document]
philium: I actually really like how having the `IBOutlet` properties after `init(coder:)` but before `viewDidLoad()` follows the flow of initialization. It is also progressive disclosure from top to bottom: the first thing you see is the model properties, then how they are initialized, then the view properties, then how those are initialized/configured. You don't see the outlets right away because they're not used right away.

Having the outlets declared with the other stored properties does keep them all together, but I actually think it is more readable the way I've broken them up. Your thoughts?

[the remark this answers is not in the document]
nixta: If we're going to separate the `init()` and `viewDidLoad()` then they should be adjacent. We need to focus on Runtime and aren't particularly interested in demonstrating UIKit/Cocoa flow.

arcgis-runtime-samples-macos/Features/Feature layer query/FeatureLayerQueryVC.swift
```diff
@@ -36,7 +36,7 @@ class FeatureLayerQueryVC: NSViewController, NSTextFieldDelegate {
         self.mapView.map = self.map
         
         //create feature table using a url
-        self.featureTable = AGSServiceFeatureTable(url: URL(string: "https://sampleserver6.arcgisonline.com/arcgis/rest/services/USA/MapServer/2")!)
```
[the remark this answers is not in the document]
philium: Thank you. I completely missed the layer specifier.

arcgis-runtime-samples-macos/Layers/Query a map image sublayer/QueryMapImageSublayerViewController.swift
philium: Thank you. I didn't know about `AGSLoadObjects()`.

arcgis-runtime-samples-macos/Layers/Query a map image sublayer/QueryMapImageSublayerViewController.swift
philium: Good question. You are right that the method could easily be inlined because it isn't called from anywhere else.

But I've found that there are a couple of practical benefits to using a separate method to respond to the change:

1. Property observers are much more readable and understandable when they are short and sweet. The longer the implementation, the more it look like a method.
2. The change handler method can be called manually. This is most useable when initializing the property to a default value as the change handler can be called from an initializer, whereas the property observer isn't fired.

So while it isn't necessary in this case, using a separate method is a design pattern I've used for quite some time.

arcgis-runtime-samples-macos/Layers/Query a map image sublayer/QueryMapImageSublayerViewController.swift
philium: I avoided using `allObjects` because of this note in the documentation:
> Accessing this property causes all the features to be loaded into memory. To conserve memory, consider iterating over the features one-by-one using a for-in loo

nixta: Stuck between a rock and a hard place 😖

Your approach is a better general pattern then, yeah.

arcgis-runtime-samples-macos/Layers/Query a map image sublayer/QueryMapImageSublayerViewController.swift
nixta: Understood about the pattern. Just wondering which is most readable in this case. This is clean the way it is (I really only brought it up because my first thought was "🤔 where else is this being called from?".

PR #109, opened 2018-06-27
Adds the Scene Layer Selection sample

philium:


arcgis-runtime-samples-macos/Scenes/Scene layer selection/SceneLayerSelectionViewController.swift
```diff
@@ -0,0 +1,94 @@
...
+            if let error = error {
+                self?.layerDidFailToLoad(with: error)
+            } else {
+                self?.layerDidLoad()
+            }
+        }
+    }
+    
+    /// The scene view managed by the view controller.
+    @IBOutlet weak var sceneView: AGSSceneView!
```
nixta: Could we move the `@IBOutlet` to the top please?

arcgis-runtime-samples-macos/Scenes/Scene layer selection/SceneLayerSelectionViewController.swift
```diff
@@ -0,0 +1,94 @@
...
+        alert.addButton(withTitle: "OK")
+        alert.beginSheetModal(for: window)
+    }
+}
+
+extension SceneLayerSelectionViewController: AGSGeoViewTouchDelegate {
+    func geoView(_ geoView: AGSGeoView, didTapAtScreenPoint screenPoint: CGPoint, mapPoint: AGSPoint) {
+        buildingsLayer.clearSelection()
+        sceneView.identifyLayer(buildingsLayer, screenPoint: screenPoint, tolerance: 10, returnPopupsOnly: false) { [weak self] (result) in
+            guard let strongSelf = self else { return }
```
nixta: Should be able to avoid the `guard`s entirely in this section with something like:

``` Swift
sceneView.identifyLayer(buildingsLayer, screenPoint: screenPoint, tolerance: 10, returnPopupsOnly: false) { [weak self] (result) in
    if let error = result.error {
        print("\(result.layerContent.name) identify failed: \(error)")
    } else if let feature = result.geoElements.first as? AGSFeature {
        self?.buildingsLayer.select(feature)
    }
}
```

I'm all for `guard` statements when they simplify the code and reduce closure depths, but I feel this is as good (or perhaps a tiny bit clearer).

philium: Thanks. my initial implementation of that closure had multiple references to `self`. I eventually simplified it but overlooked eliminating the `guard` statement.

PR #114, opened 2018-07-02
Add new sample: Cut Geometry

esreli:
This sample demonstrates how to cut a geometry with a polyline using the GeometryEngine. For example, you may want to subdivide a large parcel polygon into several smaller parcels.


arcgis-runtime-samples-macos/Geometry/Cut geometry/CutGeometry.storyboard
```diff
@@ -0,0 +1,57 @@
...
+        <scene sceneID="e7x-ur-E6P">
+            <objects>
+                <viewController id="snF-ms-OmO" customClass="CutGeometryViewController" customModule="arcgis_runtime_samples_macos" customModuleProvider="target" sceneMemberID="viewController">
+                    <view key="view" id="MPW-TP-25B">
+                        <rect key="frame" x="0.0" y="0.0" width="450" height="300"/>
+                        <autoresizingMask key="autoresizingMask"/>
+                        <subviews>
+                            <customView translatesAutoresizingMaskIntoConstraints="NO" id="aoQ-Fg-Hvg" customClass="AGSMapView">
+                                <rect key="frame" x="0.0" y="0.0" width="450" height="300"/>
+                                <subviews>
```
philium: My understanding is that we shouldn't add subviews to `AGSMapView`/`AGSSceneView` as we don't want to encourage users to do the same.

arcgis-runtime-samples-macos/Geometry/Cut geometry/CutGeometryViewController.swift
```diff
@@ -0,0 +1,150 @@
...
+import Cocoa
+import ArcGIS
+
+class CutGeometryViewController: NSViewController {
+
+    @IBOutlet weak var mapView: AGSMapView!
+    
+    let graphicsOverlay = AGSGraphicsOverlay()
+    
+    var lakeSuperiorPolygonGraphic: AGSGraphic = {
```
philium: I missed this on the iOS review. Please make `lakeSuperiorPolygonGraphic` and `borderPolylineGraphic` immutable (`let`).

arcgis-runtime-samples-macos/Geometry/Cut geometry/CutGeometry.storyboard
```diff
@@ -0,0 +1,57 @@
...
+                                    <button verticalHuggingPriority="750" translatesAutoresizingMaskIntoConstraints="NO" id="lW7-NO-G6Y">
+                                        <rect key="frame" x="355" y="252" width="81" height="32"/>
+                                        <constraints>
```
philium: @esreli You shouldn't need these constraints as `NSButton` has an intrinsic content size. You should be able to use the layout margins to adjust the padding around the title, if desired.

esreli:
@philium please re-review

arcgis-runtime-samples-macos/Geometry/Cut geometry/CutGeometryViewController.swift
```diff
@@ -0,0 +1,150 @@
+//// Copyright 2018 Esri
```
philium: Not a big deal, but I happened to notice there are two extra slashes (`/`) on this line. Was that intentional?

esreli:
@philium please re-review. There was a merge conflict with v.next in `project.pbxproj` introduced by the Scene Layer Selection PR. This should be sorted in this latest commit.

PR #115, opened 2018-07-02
Add new sample: Geodesic operations

ghost:
This sample demonstrates how to use AGSGeometryEngine to calculate a geodesic path between two points and measure its distance.


philium (dismissed):
Looks good. Thank you!

gdhakal:
@philium I have resolved merge conflicts. Can you review it please?

arcgis-runtime-samples-macos/Content Display Logic/ContentPList.plist
```diff
@@ -983,6 +983,18 @@
                     <string>GeodesicOperationsViewController</string>
                 </array>
             </dict>
+			<dict>
```
philium: The indentation doesn't look consistent with the rest of the file. You can fix it by modifying one of the elements in Xcode's plist editor. Xcode will regenerate the plist and the indentation will be consistent throughout.

philium (changes requested):

arcgis-runtime-samples-macos/Content Display Logic/ContentPList.plist
gdhakal: @philium It was a bit tricky to fix indentation esp. because the indentation looked good when opening the file in Xcode. I resorted to re-indenting the entire file; couldn't find a better way to fix it. Let me know what you think.

gdhakal: @philium Thanks for the suggestions. It worked. Please re-review.

philium (approved):
Thank you!

PR #117, opened 2018-07-03
Scene Layer Selection: Prevents crash if the layer loads before the view.

philium:
It is unlikely to happen, but it is possible.

### DIFF
--- a/arcgis-runtime-samples-macos/Scenes/Scene layer selection/SceneLayerSelectionViewController.swift
+++ b/arcgis-runtime-samples-macos/Scenes/Scene layer selection/SceneLayerSelectionViewController.swift
@@ -57,11 +57,14 @@ class SceneLayerSelectionViewController: NSViewController {
         
         let camera = AGSCamera(latitude: 48.378, longitude: -4.494, altitude: 200, heading: 345, pitch: 65, roll: 0)
         sceneView.setViewpointCamera(camera)
+        if buildingsLayer.loadStatus == .loaded {
+            sceneView.touchDelegate = self
+        }
     }
     
     /// Called in response to the layer loading successfully.
     func layerDidLoad() {
-        sceneView.touchDelegate = self
+        sceneView?.touchDelegate = self
     }
     
     /// Called in response to the layer failing to load. Presents an alert
--- a/arcgis-runtime-samples-macos/Scenes/Scene layer selection/SceneLayerSelectionViewController.swift
+++ b/arcgis-runtime-samples-macos/Scenes/Scene layer selection/SceneLayerSelectionViewController.swift
@@ -36,14 +36,6 @@ class SceneLayerSelectionViewController: NSViewController {
         scene.operationalLayers.add(buildingsLayer)
         
         super.init(coder: coder)
-        
-        buildingsLayer.load { [weak self] (error) in
-            if let error = error {
-                self?.layerDidFailToLoad(with: error)
-            } else {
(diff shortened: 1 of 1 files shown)


arcgis-runtime-samples-macos/Scenes/Scene layer selection/SceneLayerSelectionViewController.swift
```diff
@@ -57,11 +57,14 @@ class SceneLayerSelectionViewController: NSViewController {
         
         let camera = AGSCamera(latitude: 48.378, longitude: -4.494, altitude: 200, heading: 345, pitch: 65, roll: 0)
         sceneView.setViewpointCamera(camera)
+        if buildingsLayer.loadStatus == .loaded {
```
nixta: This is not how we should do this. We're just chasing status here.

The Loadable pattern was designed to prevent this kind of dual-location status checking.

But, more to the point, what's the crash? Setting touchDelegate on sceneView before it's initialized?

It looks like you don't need the `load()` at all. Just set the `touchDelegate` in `viewDidLoad` and everything else should take care of loading by AGSLoadable dependency.

nixta: In fact, this is sufficient:

``` Swift
class SceneLayerSelectionViewController: NSViewController {
    /// The scene displayed in the scene view.
    let scene: AGSScene
    let buildingsLayer: AGSArcGISSceneLayer
    
    required init?(coder: NSCoder) {
        scene = AGSScene(basemap: .imagery())
        
        // Create a surface set it as the base surface of the scene.
        let surface = AGSSurface()
        surface.elevationSources = [AGSArcGISTiledElevationSource(url: .worldElevationService)]
        scene.baseSurface = surface
        
        buildingsLayer = AGSArcGISSceneLayer(url: .brestBuildingsService)
        scene.operationalLayers.add(buildingsLayer)
        
        super.init(coder: coder)
    }
    
    /// The scene view managed by the view controller.
    @IBOutlet weak var sceneView: AGSSceneView!
    
    override func viewDidLoad() {
        super.viewDidLoad()
        
        // Assign the scene to the scene view.
        sceneView.scene = scene
        
        let camera = AGSCamera(latitude: 48.378, longitude: -4.494, altitude: 200, heading: 345, pitch: 65, roll: 0)
        sceneView.setViewpointCamera(camera)

        sceneView.touchDelegate = self
    }
}
```

This is on me for not noticing this and highlighting it sooner.

arcgis-runtime-samples-macos/Scenes/Scene layer selection/SceneLayerSelectionViewController.swift
```diff
@@ -36,14 +36,6 @@ class SceneLayerSelectionViewController: NSViewController {
         scene.operationalLayers.add(buildingsLayer)
         
         super.init(coder: coder)
-        
-        buildingsLayer.load { [weak self] (error) in
```
philium: My understanding was that the layer needed to be loaded before we could identify features. I could have used `guard buildingsLayer.loadStatus == .loaded else { return}` in the touch handler, but I wanted to avoid performing that check every single time.

I've confirmed that trying to identify before the layer is loaded simple returns an error, which we handle gracefully. Therefore I've opted to not explicitly load the layer and assign the touch delegate upon loading of the view.

PR #124, opened 2018-07-12
Refactors MapLoadedViewController to update load status before the vi…

philium:
…ew is displayed.

I also took the opportunity to cleanup the code:
• Made the map non-optional.
• Conformed the code to our style guide.


philium:
@rolson I agree about the sample not being very useful. Most of the time the map finishes loading as the map view appears, so the user only ever sees the map state as Loaded. We've talked about improving this sample by adding a button the user clicks to trigger loading the map.

I think the point of the sample is to show how to monitor all the load states, not just find out when the map finished loading. That's why it uses KVO rather than calling `load()`.

PR #125, opened 2018-07-17
Cleans up Open Existing Map sample by introducing a new type

philium:


arcgis-runtime-samples-macos/Maps/Open an existing map/OpenExistingMapViewController.swift
```diff
@@ -17,49 +17,36 @@
...
-    //MARK: - ExistingMapsListDelegate
-    
-    func existingMapsListViewController(_: ExistingMapsListViewController, didSelectItemAtIndex index: Int) {
-        var selectedPortalItemURL:String
-        switch index {
-        case 1:
-            selectedPortalItemURL = self.itemURL2
-        case 2:
-            selectedPortalItemURL = self.itemURL3
+        switch segue.destinationController {
```
nixta: Would this be better in an `if let`? Seems the switch statement, while generic and more elegant if more cases were to be added, isn't really necessary here? Just wondering if the switch statement is left over from before, or if it's the best way to do this.

philium: While an `if` statement would be a little more concise, it was a conscious decisions to use a `switch` statement. In my experience, almost all overrides of `prepare(for:sender:)` are solely interested in the destination controller. Therefore a `switch` statement works well as a general solution.

That being said, I can change it to an `if` statement if you really want me to.

nixta: If you think this is the right way to do it, let's keep it. It's not getting in the way of showing how to do anything with the Runtime.

PR #128, opened 2018-07-18
Find Address: Refactors geometry factory method

philium:
And cleans up the closure from which it is called.

### DIFF
--- a/arcgis-runtime-samples-macos/Search/Find address/FindAddressViewController.swift
+++ b/arcgis-runtime-samples-macos/Search/Find address/FindAddressViewController.swift
@@ -51,20 +51,6 @@ class FindAddressViewController: NSViewController, AGSGeoViewTouchDelegate, NSTe
         
     }
     
-    /// Creates a graphic with the specified point and attributes.
-    ///
-    /// - Parameters:
-    ///   - point: A point for the geometry of the graphic.
-    ///   - attributes: A dictionary of attributes for the graphic.
-    /// - Returns: A new `AGSGraphic` object.
-    private func makeGraphic(point: AGSPoint, attributes: [String: Any]?) -> AGSGraphic {
-        let markerImage = #imageLiteral(resourceName: "RedMarker")
-        let symbol = AGSPictureMarkerSymbol(image: markerImage)
-        symbol.leaderOffsetY = markerImage.size.height / 2
-        symbol.offsetY = markerImage.size.height / 2
-        return AGSGraphic(geometry: point, symbol: symbol, attributes: attributes)
-    }
-    
     private func geocodeSearchText(_ text:String) {
         //clear already existing graphics
         self.graphicsOverlay.graphics.removeAllObjects()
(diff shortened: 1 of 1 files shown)


arcgis-runtime-samples-macos/Search/Find address/FindAddressViewController.swift
```diff
@@ -51,15 +51,18 @@ class FindAddressViewController: NSViewController, AGSGeoViewTouchDelegate, NSTe
...
-    //also sets the leader offset and offset
-    private func graphic(for point: AGSPoint, attributes: [String: AnyObject]?) -> AGSGraphic {
-        let markerImage = NSImage(named: NSImage.Name(rawValue: "RedMarker"))!
+    /// Creates a graphic with the specified point and attributes.
+    ///
+    /// - Parameters:
+    ///   - point: A point for the geometry of the graphic.
+    ///   - attributes: A dictionary of attributes for the graphic.
+    /// - Returns: A new `AGSGraphic` object.
+    private func makeGraphic(point: AGSPoint, attributes: [String: Any]?) -> AGSGraphic {
```
pgruenler: I kind of prefer the old name, is there a reason you changed it?

philium: Good question. Since the role of this function is to manufacture graphic objects, I tend to think of it as a factory method. The [Swift API Design Guidelines](https://swift.org/documentation/api-design-guidelines/#strive-for-fluent-usage) has this to say about factory methods:

> Begin names of factory methods with “make”, e.g. x.makeIterator().

and

> The first argument to initializer and factory methods calls should not form a phrase starting with the base name, e.g. x.makeWidget(cogCount: 47)

It appears that, for the most part, the code in the samples apps follow the design guidelines. I've just been updating inconsistencies as I run across them. That's why I changed the function name.

pgruenler: I don't think we should consider this a factory method.  The Apple documentation looks to be using a pretty strict gang-of-four definition of a factory method.  This method is more of a convenience to instantiate a graphic and give it a symbol.  I think this fits more with the next section in the documentation:
>Name functions and methods according to their side-effects

>Those without side-effects should read as noun phrases, e.g. x.distance(to: y), i.successor().

Like `x.distance(to:y)` returns a distance, `x.graphic(for...` returns a graphic.

philium: @pgruenler Meaning of the term "factory method" aside, the distinction I see is the function is creating something rather than looking up something. The very first fundamental of the design guidelines is "Clarity at the point of use". Looking at the point of use of the function in question:

```swift
let graphic = strongSelf.graphic(for: results[0].displayLocation!, attributes: result.attributes)
```

It isn't entirely clear from the call site whether the function is returning an existing graphic or creating a new one. It could be creating a new graphic with the given location and attributes or it could be looking for an existing graphic at the given location with the given attributes.

Now let's look at the use of the renamed function:

```swift
let graphic = strongSelf.makeGraphic(point: result.displayLocation!, attributes: result.attributes)
```

It is clear from the call site that a new graphic is being created rather than an existing graphic being retrieved. Does that make sense?

Granted, this is a private method that is only called once, so the name of this particular function isn't that big of a deal. But as a design pattern, I've found it to be helpful to myself and other who look at the code in the future.

philium: The distinction of creating vs. retrieving is particular important in this example because the item  in question is a reference type. It is nice to know from the call site what happens when the function is called multiple times with the same arguments. Does it return the same object or distinct objects each time?

PR #129, opened 2018-07-20
Uses Implicit Member Expression for basemaps.

philium:


philium:
@pgruenler Great. Thank you!

PR #135, opened 2018-07-25
New Sample: Line of Sight (GeoElement)

nixta:


arcgis-runtime-samples-macos/Analysis/Line of sight (geoelement)/LineOfSightGeoElementViewController.swift
```diff
@@ -0,0 +1,265 @@
...
+            taxiGraphic.isSelected = false
+        }
+    }
+
+    func updateObserverZLabel() {
+        var label = "Unknown"
+        if let geom = observerGraphic.geometry as? AGSPoint {
+            label = "\(Int(geom.z))"
+        }
+        observerZLabel.stringValue = "\(label)\(observerZUnitString)"
```
philium: You might consider using a measurement formatter for setting the labels' values:

```swift

observerZLabel.stringValue = {
    if let geometry = observerGraphic.geometry as? AGSPoint {
        let measurement = Measurement<UnitLength>(value: geometry.z, unit: .meters)
        return measurementFormatter.string(from: measurement)
    } else {
        return "Unknown"
    }
}()

// Declared as a property on the view controller.
let measurementFormatter: MeasurementFormatter = {
    let formatter = MeasurementFormatter()
    formatter.numberFormatter.maximumFractionDigits = 0
    formatter.numberFormatter.roundingMode = .down
    formatter.unitOptions = .providedUnit
    return formatter
}()
```

As you know, measurement formatters are locale aware and use the current locale by default. That doesn't matter very much for the sample apps, but it does matter for user apps. It would be nice to show them how to do this the right way, if we can.

arcgis-runtime-samples-macos/Analysis/Line of sight (geoelement)/LineOfSightGeoElementViewController.swift
```diff
@@ -0,0 +1,265 @@
...
+        return AGSGeometryEngine.unionGeometries(streetIntersectionLocations)?.extent.center ?? observerPoint
+    }
+
+    required init?(coder aDecoder: NSCoder) {
+        // ====================================
+        // set up the scene, layers and overlay
+        // ====================================
+
+        // initialize the scene with an imagery basemap
+        scene = AGSScene(basemap: AGSBasemap.imageryWithLabels())
```
philium: Please remove `AGSBasemap`. All of the other samples now use implicit member references for the basemap wherever possible.

arcgis-runtime-samples-macos/Analysis/Line of sight (geoelement)/LineOfSightGeoElementViewController.swift
```diff
@@ -0,0 +1,284 @@
...
+        sceneView.graphicsOverlays.add(overlay)
+
+        // create an analysis overlay using a single Line of Sight and add it to the scene view
+        let analysisOverlay = AGSAnalysisOverlay()
+        analysisOverlay.analyses.add(lineOfSight)
+        sceneView.analysisOverlays.add(analysisOverlay)
+
+        // update the UI if the Line of Sight analysis result changes
+        losObserver = lineOfSight.observe(\.targetVisibility, options: .new) { (losAnalysis, _) in
+            DispatchQueue.main.async { [weak self] in
```
philium: I thought for sure I mentioned this in my first review, but apparently not. Please move the capture list to the outer closure. The way it is now, the outer closure is strongly capturing `self`.

PR #140, opened 2018-07-31
Merge branch release/100.3 into master

philium:


arcgis-runtime-samples-macos/Features/Feature layer query/FeatureLayerQueryVC.swift
```diff
@@ -36,7 +36,7 @@ class FeatureLayerQueryVC: NSViewController, NSTextFieldDelegate {
         self.mapView.map = self.map
         
         //create feature table using a url
-        self.featureTable = AGSServiceFeatureTable(url: URL(string: "https://sampleserver6.arcgisonline.com/arcgis/rest/services/USA/MapServer/2")!)
+        self.featureTable = AGSServiceFeatureTable(url: URL.unitedStatesMapService.appendingPathComponent("2"))
```
dg0yal: ugh, i don't like this. Takes something very simple to read and obfuscates it!

philium: @dg0yal I understand. The initial thought was to not have the base URL in multiple places. If the map is moved to a different URL in the future, it would only need to be changed in one location. 

That being said, there are different ways of achieving the same result. We brainstormed a couple different ways at the time this change was made.

PR #142, opened 2018-08-03
Cleans up SurfacePlacementsViewController.

philium:
Refactors sample to be more straightforward and adds documentation comments.

### DIFF
--- a/arcgis-runtime-samples-macos/Scenes/Surface placements/SurfacePlacementsViewController.swift
+++ b/arcgis-runtime-samples-macos/Scenes/Surface placements/SurfacePlacementsViewController.swift
@@ -46,7 +46,11 @@ class SurfacePlacementsViewController: NSViewController {
         let camera = AGSCamera(latitude: 53.04, longitude: -4.04, altitude: 1300, heading: 0, pitch: 90, roll: 0)
         self.sceneView.setViewpointCamera(camera)
         
-        let graphicsOverlays = AGSSurfacePlacement.allCases.map { makeGraphicsOverlay(surfacePlacement: $0) }
+        let graphicsOverlays = [
+            makeGraphicsOverlay(surfacePlacement: .draped),
+            makeGraphicsOverlay(surfacePlacement: .relative),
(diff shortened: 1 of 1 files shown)


arcgis-runtime-samples-macos/Scenes/Surface placements/SurfacePlacementsViewController.swift
```diff
@@ -17,58 +17,71 @@
...
-        self.drapedGraphicsOverlay.sceneProperties?.surfacePlacement = .draped
-        self.relativeGraphicsOverlay.sceneProperties?.surfacePlacement = .relative
-        self.absoluteGraphicsOverlay.sceneProperties?.surfacePlacement = .absolute
-        
-        //add graphic overlays to the scene view
-        self.sceneView.graphicsOverlays.addObjects(from: [self.drapedGraphicsOverlay, self.relativeGraphicsOverlay, self.absoluteGraphicsOverlay])
-        
-        //add graphics
-        self.addGraphics()
+        let graphicsOverlays = AGSSurfacePlacement.allCases.map { makeGraphicsOverlay(surfacePlacement: $0) }
```
nixta: While this is nice and Swifty, I wonder if the sample would benefit from three separate lines calling `makeGraphicsOverlay()` with each `AGSSurfacePlacement` specified explicitly.

I think this is very elegant, but makes the reader (especially one who isn't very familiar with Swift) dig too deep to get to the crux of things, namely that we have `.draped`, `.relative` and `.absolute`.

Perhaps:

``` Swift
sceneView.graphicsOverlays.add(makeGraphicsOverlay(surfacePlacement: .draped))
sceneView.graphicsOverlays.add(makeGraphicsOverlay(surfacePlacement: .relative))
sceneView.graphicsOverlays.add(makeGraphicsOverlay(surfacePlacement: .absolute))
```

philium: Good idea. I'll change it.

PR #145, opened 2018-08-07
 Refactors MainViewController into separate view controllers.

philium:
Using proper view controller containment. Everything should work the way it did before, more or less. I made a few tweaks here and there:

- The enclosing category is now expanded when a sample is selected.
- Search no longer returns an incorrect number of results the first time.
- The divider is now moveable as a result of using a split view controller.
- Sample child view controller views are only loaded and displayed as needed as a result of using a tab view controller.


arcgis-runtime-samples-macos/Content Display Logic/Controllers/MainViewController.swift
```diff
@@ -40,296 +34,85 @@ extension NSView {
...
-        self.clearPlaceholderView()
-        self.collectionViewController.view.frame = self.placeholderView.bounds
-        self.collectionViewController.sampleNodes = sampleNodes
-        self.placeholderView.addSubview(self.collectionViewController.view)
+}
+
+extension MainViewController: SampleCollectionViewControllerDelegate {
+    func sampleCollectionViewController(_ controller: SampleCollectionViewController, didSelect sample: Node) {
+        let sampleListViewController = sampleListSplitViewItem.viewController as! SampleListViewController
+        sampleListViewController.select(sample)
```
philium: I was surprised to find that selecting a row in code resulted in the delegate method being called. I would expect it to only happen as a result of user action. Regardless, the sample is actually displayed as a result of the selection in the sample list.

arcgis-runtime-samples-macos/Content Display Logic/Controllers/ReadmeViewController.swift
```diff
@@ -19,12 +19,11 @@ import WebKit
...
-        didSet {
-            if self.folderName != nil {
-                self.fetchFileContent(self.folderName)
-            }
-        }
+    var folderName = ""
+    
+    override func viewDidLoad() {
+        super.viewDidLoad()
+        fetchFileContent(folderName)
```
philium: This was moved here because now the view has not been loaded when `folderName` is set.

arcgis-runtime-samples-macos/Content Display Logic/Controllers/SourceCodeViewController.swift
```diff
@@ -21,24 +21,26 @@ class SourceCodeViewController: NSViewController, NSSearchFieldDelegate {
...
-            //populate popUpButton
-            self.popUpButton.addItems(withTitles: self.fileNames)
+    var fileNames = [String]()
+    
+    override func viewDidLoad() {
+        super.viewDidLoad()
+        
+        //load the source code
+        if let fileName = fileNames.first {
+            loadHTMLPage(fileName)
```
philium: This was moved here because now the view has not been loaded when `fileNames` is set.

arcgis-runtime-samples-macos/Content Display Logic/Controllers/WindowController.swift
```diff
@@ -33,7 +33,8 @@ extension NSWindow {
 }
 
 class WindowController: NSWindowController, NSSearchFieldDelegate, NSWindowDelegate, SuggestionsVCDelegate {
-
+    let searchEngine = SearchEngine()
```
philium: Creating this here gives some time for the engine to load before the user executes their first query. This fixes a problem where the results were incorrect for the first search.

arcgis-runtime-samples-macos.xcodeproj/project.pbxproj
```diff
@@ -815,6 +821,10 @@
...
 		4CE3B22A20E2AD2000D4A06A /* SceneLayerSelectionViewController.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = SceneLayerSelectionViewController.swift; sourceTree = "<group>"; };
 		4CE3B22C20E2AD2D00D4A06A /* SceneLayerSelection.storyboard */ = {isa = PBXFileReference; lastKnownFileType = file.storyboard; path = SceneLayerSelection.storyboard; sourceTree = "<group>"; };
+		4CF0258021151F9C00806810 /* SampleViewController.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = SampleViewController.swift; sourceTree = "<group>"; };
+		4CF92E8F2118BB97009A62C6 /* SampleCollectionViewController.xib */ = {isa = PBXFileReference; lastKnownFileType = file.xib; path = SampleCollectionViewController.xib; sourceTree = "<group>"; };
+		4CF92E942118D9FF009A62C6 /* SampleCollectionHeaderView.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = SampleCollectionHeaderView.swift; sourceTree = "<group>"; };
+		4CF92E962118DABE009A62C6 /* SampleCollectionHeaderView.xib */ = {isa = PBXFileReference; lastKnownFileType = file.xib; path = SampleCollectionHeaderView.xib; sourceTree = "<group>"; };
```
nixta: Is there a reason this needs to be in a separate file rather than contained and controlled by the SampleCollectionViewController xib and code-behind?

arcgis-runtime-samples-macos/Content Display Logic/Sample Collection View/SampleCollectionHeaderView.swift
```diff
@@ -0,0 +1,25 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+import Cocoa
+
+class SampleCollectionHeaderView: NSView {
```
nixta: I can see that this makes sense to have as a separate swift file, but does it need a separate xib? Strikes me this is part of the overall SampleCollectionView hierarchy.

arcgis-runtime-samples-macos.xcodeproj/project.pbxproj
philium: Good question. I've always thought that a cell registered using `NSCollectionView.register(_:forSupplementaryViewOfKind:withIdentifier:)` had to be in a xib of its own. Maybe it doesn't? I'll try and see.

philium: And that thought was correct. From the documentation for the referenced `NSCollectionView` method:

> The nib file must contain exactly one NSView object at the top level and that view must conform to the NSCollectionViewElement protocol.

philium:
@nixta I tried to use the main storyboard as much as possible. But due to the dynamic nature of our UI, it wasn't possible to model everything in a storyboard. The view controller in the right split view item is based off of the selection in the left split view item and there isn't really a good way to model that in a storyboard. The iOS workflow is much easier to model in a storyboard: Categories > Samples > Sample.

nixta:
I do that in the maps-app-ios, though granted there's some code that needs to go with it to swap out subviews, so fair enough.

However, I still don't see why the SampleCollectionHeaderView needs to be a separate XIB? Shouldn't that just be a subview in SampleCollectionViewController.xib's view hierarchy?

nixta:
Sorry. I didn't realize/notice that the SampleCollectionHeaderView was a collection view section heading. Looks good then.

philium:
@nixta @qsmmorgan I pushed a change that dismissed your approvals. Please re-review.

PR #153, opened 2018-08-16
Adds the Feature Layer Rendering Mode (Map) sample

philium:


esreli:
screen shot 2018-08-21 at 2 01 48 pm

The left pane does not allow the user to click drag to reveal the full title of the sample. This means feature titles are permanently trimmed.

arcgis-runtime-samples-macos/Features/Feature layer rendering mode (map)/FeatureLayerRenderingModeMapViewController.swift
```diff
@@ -0,0 +1,162 @@
...
+        switch zoomState {
+        case .zoomedIn, .zoomingIn:
+            zoomOut(animated: animated)
+        case .zoomedOut, .zoomingOut:
+            zoomIn(animated: animated)
+        }
+    }
+    
+    /// Animates the map to the zoomed in viewpoint.
+    func zoomIn(animated: Bool) {
```
esreli: It feels as though there's a lot of repeat, conditional code between this function and `zoomOut`. Could you consolidate the two?

philium:
@esreli Thank you for pointing that out. I have a [separate PR](https://github.com/Esri/arcgis-runtime-samples-macos/pull/145) that, among other things, fixes that issue.

arcgis-runtime-samples-macos/Features/Feature layer rendering mode (map)/FeatureLayerRenderingModeMapViewController.swift
```diff
@@ -0,0 +1,162 @@
...
+        let point = AGSPoint(x: -118.45, y: 34.395, spatialReference: .wgs84())
+        let scale = scaleRange.lowerBound
+        let viewpoint = AGSViewpoint(center: point, scale: scale, rotation: 90)
+        setViewpoint(viewpoint, of: staticMapView, animated: animated)
+        setViewpoint(viewpoint, of: dynamicMapView, animated: animated) { [weak self] (finished) in
+            guard finished else { return }
+            self?.zoomState = .zoomedIn
+        }
+        // Update the zoom button.
+        zoomButton.title = "Zoom Out"
```
esreli: The title could be a computed member of `ZoomState`

arcgis-runtime-samples-macos/Features/Feature layer rendering mode (map)/FeatureLayerRenderingModeMapViewController.swift
```diff
@@ -0,0 +1,162 @@
...
+    }
+    
+    /// Animates the map to the zoomed in viewpoint.
+    func zoomIn(animated: Bool) {
+        // Set the viewpoint of the map views.
+        let point = AGSPoint(x: -118.45, y: 34.395, spatialReference: .wgs84())
+        let scale = scaleRange.lowerBound
+        let viewpoint = AGSViewpoint(center: point, scale: scale, rotation: 90)
+        setViewpoint(viewpoint, of: staticMapView, animated: animated)
+        setViewpoint(viewpoint, of: dynamicMapView, animated: animated) { [weak self] (finished) in
```
esreli: Would you consider using a `DispatchQueue`? This would ensure both view point animation callback have completed instead of assuming they both have based on the completion of one of them.

arcgis-runtime-samples-macos/Features/Feature layer rendering mode (map)/FeatureLayerRenderingModeMapViewController.swift
```diff
@@ -0,0 +1,162 @@
...
+    /// Changes the viewpoint of the given mapview.
+    ///
+    /// - Parameters:
+    ///   - viewpoint: The viewpoint to be set.
+    ///   set.
+    ///   - mapView: The mapview whose viewpoint should be set.
+    ///   - animated: Specify `true` if the viewpoint change should be animated.
+    ///   - completion: A closure to execute after the animation has finished.
+    func setViewpoint(_ viewpoint: AGSViewpoint, of mapView: AGSMapView, animated: Bool, completion: ((Bool) -> Void)? = nil) {
+        if animated {
```
esreli: Is this necessary? If animated is false, the duration could be set to 0. Perhaps you might consider:

`let duration: TimeInterval = animated ? 5 * deltaScale / (scaleRange.upperBound - scaleRange.lowerBound) : 0`

arcgis-runtime-samples-macos/Features/Feature layer rendering mode (map)/FeatureLayerRenderingModeMapViewController.swift
esreli: Further, you could apply a `didSet` closure on `zoomState` that updates `zoomButton.title`

arcgis-runtime-samples-macos/Features/Feature layer rendering mode (map)/FeatureLayerRenderingModeMapViewController.swift
esreli: I would add an additional enum:

```swift
enum ZoomDirection {
    
    case in
    case out

    var point { get }
    var scale { get }
    var rotation { get }
    var viewpoint { get }
}
```

which could consolidate zoom:

```swift
func zoom(_ direction: ZoomDirection, animated: Bool) {
    let viewpoint = direction.viewpoint
    setViewpoint(viewpoint ...
    ...
}

...

zoom(.in, animated: true)
```

arcgis-runtime-samples-macos/Features/Feature layer rendering mode (map)/FeatureLayerRenderingModeMapViewController.swift
esreli: _Note_, above I originally wrote `DispatchQueue` when I meant to write `DispatchGroup`. I have updated the comment.

arcgis-runtime-samples-macos/Features/Feature layer rendering mode (map)/FeatureLayerRenderingModeMapViewController.swift
philium: Thank you for the suggestion. 

At first glance it sounds like that might be a good idea, but it would actually end up increasing the amount of code without any benefit. For example, consider the four proposed properties: `point`, `scale`, `rotation`, and `viewpoint`. Each of them would need to switch over `self`, increasing the number of switch statements from one (in `zoom(_:)` to five.

I do agree that it feels as though there is repetition between `zoomIn(animated:)` and `zoomOut(animated:)`, but what is being repeated is the structure, not the code itself. There is symmetry between the two functions.

The two lines that are somewhat identical are the setting of the viewpoints. That could be extracted into a separate function, something like: 

```swift
func setViewpoints(to viewpoint: AGSViewpoint, animated: Bool, completion: @escaping (Bool) -> Void) {
    setViewpoint(viewpoint, of: staticMapView, animated: animated)
    setViewpoint(viewpoint, of: dynamicMapView, animated: animated, completion: completion)
}
```

And that could possibly be combined with `setViewpoint(_:of:animated:completion:)`. I toyed with that for a bit, but in the end it seemed like more hassle than it was worse. It didn't really save any lines of code and ended up being less readable.

arcgis-runtime-samples-macos/Features/Feature layer rendering mode (map)/FeatureLayerRenderingModeMapViewController.swift
philium: I thought about that, but what would be the benefit? In either case it would add an extra switch statement. And for what? Readability? I'm not sure it makes it any more readable.

arcgis-runtime-samples-macos/Features/Feature layer rendering mode (map)/FeatureLayerRenderingModeMapViewController.swift
philium: I hadn't considered using a `DispatchGroup`. That is an interesting idea, but I'm not sure the overhead is needed. What matters is that the completion handler gets called once, not necessarily that it gets called after both animations finish. I did pick the dynamic map view as it is more likely fo finish last, but even if it didn't, it wouldn't matter. If the other animation hasn't finished, it will get interrupted and the new animation will start.

arcgis-runtime-samples-macos/Features/Feature layer rendering mode (map)/FeatureLayerRenderingModeMapViewController.swift
philium: I think in this case using the ternary operator would sacrifice readability for conciseness. I'd like to keep the `if` statement. That being said, I could adopt your other suggestion if using `0` for the animation duration when `animated` is `false`:

```swift
func setViewpoint2(_ viewpoint: AGSViewpoint, of mapView: AGSMapView, animated: Bool, completion: ((Bool) -> Void)? = nil) {
    let duration: TimeInterval
    if animated {
        // Determine the duration based on how close the map view's current
        // scale is to the target scale.
        let deltaScale = abs(viewpoint.targetScale - mapView.mapScale)
        duration = 5 * deltaScale / (scaleRange.upperBound - scaleRange.lowerBound)
    } else {
        duration = 0
    }
    mapView.setViewpoint(viewpoint, duration: duration, completion: completion)
}
```

But is that any better? I'm not sure. It is the exact same number of lines of code.

arcgis-runtime-samples-macos/Features/Feature layer rendering mode (map)/FeatureLayerRenderingModeMapViewController.swift
esreli: It would tightly couple the button title to the zoomState.

arcgis-runtime-samples-macos/Features/Feature layer rendering mode (map)/FeatureLayerRenderingModeMapViewController.swift
esreli: I don't think you would disagree it is a bit more _swifty_ to implement a `ZoomDirection` enum. Because both zoom direction cases are concrete, using an enum is most perfect, except for the number of lines of code (because switch statements are ugly!). 

You could also consider building `ZoomDirection` as a struct:

```swift
    struct ZoomDirection {
        let scale: Double
        let point: AGSPoint
        let rotation: Double
        var viewpoint: AGSViewpoint {
            return AGSViewpoint(center: point, scale: scale, rotation: rotation)
        }
    }
    
    let zoomIn: ZoomDirection = {
        return ZoomDirection(scale: 50_000.0, point: AGSPoint(x: -118.45, y: 34.395, spatialReference: .wgs84()), rotation: 90)
    }()
    
    let zoomOut: ZoomDirection = {
        return ZoomDirection(scale: 650_000.0, point: AGSPoint(x: -118.37, y: 34.46, spatialReference: .wgs84()), rotation: 0)
    }()
```

Which would allow you to consolidate following a similar design pattern as suggested previously, though you would have to make static the `lowerBounds` and `upperBounds` of `scaleRange` as well as maintain the completed `zoomState` value.

When I see repeated lines of code, or what feels like repeated lines of code, I always try to consolidate.

arcgis-runtime-samples-macos/Features/Feature layer rendering mode (map)/FeatureLayerRenderingModeMapViewController.swift
esreli: I don't think you would disagree using a `DispatchGroup` is best practice, no? Sure, there is overhead but can you suggest a scenario where that overhead is actually felt? I can't. If you are informing state, I would like to see that state accurately reflected.

arcgis-runtime-samples-macos/Features/Feature layer rendering mode (map)/FeatureLayerRenderingModeMapViewController.swift
philium: I could add a `title` property to `ZoomState`, but that would be rather strange:
```swift
extension ZoomState {
    var title: String {
        switch self {
            case .zoomingIn, .zoomedIn: return "Zoom Out"
            case .zoomingOut, .zoomedOut: return "Zoom In"
        }
    }
}

ZoomingState.ZoomedIn.title // "Zoomed Out"
```
The title of the `zoomedIn` state is "Zoomed Out"? That doesn't make much sense. Why? Because the value of `title` isn't actually the title of the *current* state.

Does the button title need to be updated every time the state changes? No. It only needs to be updated when changing direction, which is exactly what happens when the button Is clicked. Therefore updating the buttons title in `zoomIn(animated:)` and `zoomOut(animated:)`, which are called in response to the button being clicked, makes more sense than updating them in a `didSet` property observer of `zoomState`.

arcgis-runtime-samples-macos/Features/Feature layer rendering mode (map)/FeatureLayerRenderingModeMapViewController.swift
```diff
@@ -0,0 +1,161 @@
...
+        zoomState = .zoomingIn
+    }
+    
+    /// Animates the map to the zoomed out viewpoint.
+    func zoomOut(animated: Bool) {
+        // Set the viewpoint of the map views.
+        let point = AGSPoint(x: -118.37, y: 34.46, spatialReference: .wgs84())
+        let viewpoint = AGSViewpoint(center: point, scale: scaleRange.upperBound, rotation: 0)
+        setViewpoint(viewpoint, of: staticMapView, animated: animated)
+        setViewpoint(viewpoint, of: dynamicMapView, animated: animated) { [weak self] (finished) in
```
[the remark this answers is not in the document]
philium: What practical difference would adding a `DispatchGroup` make? The sample would behave exactly the same either way. And I'm not assuming that "the dynamicMapView animation is complete because it has been assigned the same duration." On the contrary, I calculate the duration separately for each map view. I assume that they are not totally in sync and adjust the animation duration accordingly.

arcgis-runtime-samples-macos/Features/Feature layer rendering mode (map)/FeatureLayerRenderingModeMapViewController.swift
philium: I agree that it is a bit more concise, but it isn't necessarily better for two reasons:
- `duration` is now mutable.
- `duration` is initialized twice when `animated` is `true`.

arcgis-runtime-samples-macos/Features/Feature layer rendering mode (map)/FeatureLayerRenderingModeMapViewController.swift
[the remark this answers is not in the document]
esreli: Practically speaking the app updates it's state and zoomButton title. It would be more accurate this way.

arcgis-runtime-samples-macos/Features/Feature layer rendering mode (map)/FeatureLayerRenderingModeMapViewController.swift
philium: A struct is certainly better than an enumeration. But repetition, in and of itself, isn't necessarily a bad thing. Since we only have two states, the `ZoomDirection` abstraction would actually take up more lines of code than it would save. If we had more animations, abstracting the animation properties into something like a `ZoomAnimation` type might be more beneficial.

Could I have used an abstraction in the first place? Sure. Would it have been better? Maybe. Is it worth changing now? No. The way it is now isn't *bad*, in my humble opinion. And changing it wouldn't affect the demonstration of static versus dynamic rendering, which is the whole point of the sample.

PR #160, opened 2018-09-10
Updates gitignore to match iOS

philium:
This fixes a problem where shared files inside the project were not being tracked.


qsmmorgan:
Oops! Missed that `Xcode` line

PR #162, opened 2018-09-11
Renames the "Open an existing map" sample to "Open map (URL)"

qsmmorgan:
Also replaces the map list table view with a popup button


arcgis-runtime-samples-macos/Content Display Logic/ContentPList.plist
```diff
@@ -1039,7 +1039,7 @@
 				<key>storyboardName</key>
 				<string>ListTransformations</string>
 				<key>descriptionText</key>
-				<string>This sample demonstrates how to use the transformation catalog to get a list of available datum transformations that can be used to project a geometry between two different spatial references, and how to use one of the transformations to perform the geometry engine's "project" operation.</string>
+				<string>This sample demonstrates how to use the transformation catalog to get a list of available datum transformations that can be used to project a geometry between two different spatial references, and how to use one of the transformations to perform the geometry engine&apos;s &quot;project&quot; operation.</string>
```
pgruenler: The quotes changed to - `&apos;` and `&quot;`  Is that correct?

arcgis-runtime-samples-macos/Maps/Open map (URL)/README.md
```diff
@@ -1,6 +1,6 @@
-# Open an existing map
+# Open a map from a URL
```
philium: This should be the name of the sample: Open map (URL)

qsmmorgan: Fixed in the latest commit

arcgis-runtime-samples-macos/Content Display Logic/ContentPList.plist
```diff
@@ -565,18 +563,18 @@
...
-                <key>storyboardName</key>
-                <string>FeatureLayerRenderingModeMap</string>
-                <key>descriptionText</key>
-                <string>Control the rendering mode of feature layers.</string>
-                <key>sourceFileNames</key>
-                <array>
-                    <string>FeatureLayerRenderingModeMapViewController</string>
-                </array>
-            </dict>
+			<dict>
```
philium: When I notice that Xcode wants to make other changes to this file, I usually submit those changes in a separate PR. It can help the reviewer of your changes and also the reviewer of the non-functional changes Xcode wants to make. You might consider doing that in the future.

arcgis-runtime-samples-macos/Maps/Open map (URL)/OpenMapURLViewController.swift
```diff
@@ -0,0 +1,61 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+import Cocoa
+import ArcGIS
+
+class OpenMapURLViewController: NSViewController {
+    
+    struct MapAtURL {
```
philium: I'm not sure this type is even useful anymore. It would be nice to create and configure the menu items in Interface Builder. We could then have a separate method for each map (`showHousingWithMortgages()`, etc). which would create a map with the appropriate URL and assign it to the map view.

What do you think?

arcgis-runtime-samples-macos/Content Display Logic/ContentPList.plist
qsmmorgan: Okay, I'll try to do that. In this case the inclusion of the automated changes was accidental. I usually discard them.

arcgis-runtime-samples-macos/Maps/Open map (URL)/OpenMapURLViewController.swift
qsmmorgan: I'm of two minds. My personal preference is to keep the model and UI as separate as possible. I also think it's better to have the complete map model in one place, rather than split with name/image in the storyboard and the URL in code. However, I know that some people prefer interface builder. If we go that route, I would prefer a single function that grabs the URL from an array using the `indexOfSelectedItem` property of `NSPopUpButton` rather than different functions for each item.

qsmmorgan: Ideally I would suggest getting the map names and thumbnails from the remote API, but I'm not sure if that's possible or compatible with what the other samples do.

PR #168, opened 2018-09-14
Converts the table in the ArcGIS vector tiled layer (custom style) sample to a popup

qsmmorgan:


arcgis-runtime-samples-macos/Layers/ArcGIS vector tiled layer (custom style)/VectorTileCustomStyleVC.swift
```diff
@@ -17,59 +17,108 @@
...
+            // set the basemap for the selected item
+            setBasemap(item: item)
         }
-        let controller = segue.destinationController as! VectorStylesViewController
-        controller.delegate = self
     }
+}
+
+/// A model for the items in the vectors styles popup menu
+private struct VectorStyleItem {
```
philium: Consider making `VectorStyleItem` a subtype of `VectorTileCustomStyleVC`, but keeping the interface as lean as possible by defining `itemURL` and `thumbnailImage` in an extension:
```swift
class VectorTileCustomStyleVC: NSViewController {
    …
    struct VectorStyleItem {
        var itemId: String
        var label: String
        var color1: NSColor
        var color2: NSColor
    }
    …
}

extension VectorTileCustomStyleVC.VectorStyleItem {
    var itemURL: URL { … }
    var thumbnailImage: NSImage { … }
}
```

qsmmorgan: Good idea! I didn't know you could extend subtypes like `extension VectorTileCustomStyleVC.VectorStyleItem`.

PR #171, opened 2018-09-18
Updates the WMS layer (URL) sample with a new endpoint

qsmmorgan:


qsmmorgan:
Thanks for the feedback. The README file doesn't currently display the images in the app itself, so I just used the largest thumbnail image since not many people will see it. I can still take a larger screenshot.

gdhakal:
@qsmmorgan you're right but it shows in [GitHub repo](https://github.com/Esri/arcgis-runtime-samples-macos/tree/9391eb346ab4414cec76a8bccca2939b7842797d/arcgis-runtime-samples-macos/Layers/WMS%20layer%20(URL)).

qsmmorgan:
Okay, I added the new image.

PR #172, opened 2018-09-18
Re-adds and refactors the progress indicator

qsmmorgan:
This PR re-adds the indeterminate `NSProgressIndicator` to the main window's toolbar and changes it to the correct color.

This PR also moves `hideProgressIndicator` and `showProgressIndicator` to an extension of `NSApplication` rather than `NSWindow`. This provides a number of advantages:

- It implies the global nature of the indicator.
- Calls can be made cleanly via `NSApp` rather than finding the main window on every call.
- It works in cases where a view or its window property may be `nil`, such as during `viewDidLoad` or in a completion closure where the view has already been deallocated.

### DIFF
--- a/arcgis-runtime-samples-macos/Content Display Logic/Controllers/WindowController.swift
+++ b/arcgis-runtime-samples-macos/Content Display Logic/Controllers/WindowController.swift
@@ -16,19 +16,17 @@
 
 import Cocoa
 
-extension NSWindow {
+extension NSApplication {
     //MARK: - Progres indicator
     
     func showProgressIndicator() {
-        if let controller = self.windowController as? WindowController {
-            controller.showProgressIndicator()
-        }
+        let controller = windows.compactMap { return $0.windowController as? WindowController }.first
+        controller?.showProgressIndicator()
     }
     
     func hideProgressIndicator() {
-        if let controller = self.windowController as? WindowController {
-            controller.hideProgressIndicator()
-        }
+        let controller = windows.compactMap { return $0.windowController as? WindowController }.first
+        controller?.hideProgressIndicator()
     }
 }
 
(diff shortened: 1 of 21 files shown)


arcgis-runtime-samples-macos/Content Display Logic/Controllers/WindowController.swift
```diff
@@ -16,19 +16,17 @@
...
 
-extension NSWindow {
+extension NSApplication {
     //MARK: - Progres indicator
     
     func showProgressIndicator() {
-        if let controller = self.windowController as? WindowController {
-            controller.showProgressIndicator()
-        }
+        let controller = windows.compactMap { return $0.windowController as? WindowController }.first
```
philium: As of Swift 4.2, a better solution would be to use `Collection.firstIndex(where:)`. It doesn't matter much in this case because the `windows` collection is so small, but the advantage over `compactMap` is that it doesn't need to traverse the entire collection.
```swift
var windowController: WindowController? {
    if let index = windows.firstIndex(where: { $0 is WindowController }) {
        return windows[index] as! WindowController
    } else {
        return nil
    }
}
```

qsmmorgan: I thought there must be a function like that! Thanks, I'll use that in the future.

philium: Although that's not much different than using `Collection.first(where:)`, which works prior to Swift 4.2:
```swift
var windowController: WindowController? {
    if let window = windows.first(where: { $0 is WindowController }) {
        return window as! WindowController
    } else {
        return nil
    }
}
```

qsmmorgan: I know, but then you have to cast anyway.

qsmmorgan: Oh I guess I would use `first` over `firstIndex`. I just want a single function that does the search and cast at the same time haha.

PR #173, opened 2018-09-18
Updates the statistical query (group and sort) sample to use a new service

qsmmorgan:
The thumbnail image would be nearly identical so I didn't feel the need to update it.


arcgis-runtime-samples-macos/Analysis/Statistical query (group and sort)/StatisticalQueryGroupAndSortViewController.swift
```diff
@@ -43,8 +43,10 @@ class StatisticalQueryGroupAndSortViewController: NSViewController, NSTableViewD
     override func viewDidLoad() {
         super.viewDidLoad()
         
+        let serviceURL = URL(string: "https://services.arcgis.com/jIL9msH9OI208GCb/arcgis/rest/services/Counties_Obesity_Inactivity_Diabetes_2013/FeatureServer/0")!
```
esreli: Service URLs are stored statically as a Swift Extension. Is there a reason this URL doesn't follow the same pattern? Can you follow the same pattern as the other static urls?

qsmmorgan: Inlining the URL makes it visible in the source code viewer within the app, so the user doesn't have to open Xcode to figure out what the endpoint is.

qsmmorgan: I do think it's best to keep each sample as self-contained as possible. I would prefer to inline all the URL extensions unless there is some reasoning I'm not aware of.

esreli: Sure thing. My only request is a uniformed approach.

esreli: Since I am not a member of the SDK team, i'll leave this decision up them and back out of this review.

PR #174, opened 2018-09-18
Improves sample loading

philium:
This replaces the Node class with the Category and Sample value types.

Corresponding iOS PR: https://github.com/Esri/arcgis-runtime-samples-ios/pull/317

### DIFF
--- a/arcgis-runtime-samples-macos/Content Display Logic/Controllers/SampleListViewController.swift
+++ b/arcgis-runtime-samples-macos/Content Display Logic/Controllers/SampleListViewController.swift
@@ -16,80 +16,184 @@
 
 import AppKit
 
+/// The protocol you implement to respond as the user interacts with the
+/// document browser.
 protocol SampleListViewControllerDelegate: AnyObject {
-    func sampleListViewController(_ controller: SampleListViewController, didSelect node: Node)
+    /// Called in response to the sample list selection changing.
+    ///
+    /// - Parameter controller: The current sample list.
+    func sampleListViewControllerSelectionDidChange(_ controller: SampleListViewController)
 }
 
+/// A view controller that manages an interface for displaying a list of samples
+/// by category.
 class SampleListViewController: NSViewController {
-    var nodes = [Node]() {
+    /// The delegate of the view controller.
+    weak var delegate: SampleListViewControllerDelegate?
+    /// The categories displayed by the view controller.
+    var categories = [Category]() {
         didSet {
-            guard isViewLoaded else { return }
-            outlineView.reloadData()
-            select(nodes.first!)
+            categoriesDidChange()
         }
     }
-    weak var delegate: SampleListViewControllerDelegate?
     
-    @IBOutlet weak var outlineView: NSOutlineView!
-    
-    func select(_ node: Node) {
-        // Expand the parent (if there is one).
-        if let parentNode = parent(for: node) {
-            outlineView.expandItem(parentNode)
-        }
-        // Select the new item.
-        let index = outlineView.row(forItem: node)
+    /// Selects the given category in the list.
+    ///
+    /// - Parameter category: The category to select.
+    func select(_ category: Category) {
+        guard isViewLoaded else { return }
+        guard let boxedCategory = boxedCategories[category] else { return }
+        let index = outlineView.row(forItem: boxedCategory)
         outlineView.selectRowIndexes([index], byExtendingSelection: false)
     }
     
-    /// Returns the parent of a given node.
-    ///
-    /// - Parameter node: The node whose parent should be found.
-    /// - Returns: The node's parent or `nil` if the node does not have a parent.
-    func parent(for node: Node) -> Node? {
+    /// The category currently selected in the list or `nil` if no category
+    /// is selected.
+    var selectedCategory: Category? {
+        guard isViewLoaded else { return nil }
         let selectedRow = outlineView.selectedRow
-        if selectedRow != -1, let selectedNode = outlineView.item(atRow: selectedRow) as? Node, selectedNode.childNodes.contains(node) {
-            return selectedNode
+        if selectedRow != -1, let selectedBox = outlineView.item(atRow: selectedRow) as? Box<Category> {
+            return selectedBox.value
         } else {
-            // This is a major hack. If the parent node is not the selected
-            // node, assume that the node was a selected search result. In that
-            // case, we want to skip the "Featured" category and return the
-            // sample's actual category.
-            return nodes.dropFirst().first(where: { $0.childNodes.contains(node) })
+            return nil
         }
     }
+    
+    /// Selects the given sample in the list. This will expand the sample's
+    /// category if it isn't already.
+    ///
+    /// - Parameter sample: The sample to select.
+    func select(_ sample: Sample) {
+        guard isViewLoaded else { return }
+        guard let boxedSample = boxedSamples[sample] else { return }
+
+        let row = outlineView.row(forItem: boxedSample)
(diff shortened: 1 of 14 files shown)


arcgis-runtime-samples-macos/Content Display Logic/Controllers/SampleListViewController.swift
```diff
@@ -16,80 +16,184 @@
...
+        select(categories.first!)
+    }
+    
+    // Why manually box categories and samples? Good question.
+    // While Swift would box them automatically, it would box into a new
+    // instance each time. Because NSOutlineView compares items by their memory
+    // address, two boxed instances of the same value would never be equal.
+    // By boxing ourselves, we ensure that item comparison actually works.
+    //
+    // Why not make Category and Sample reference types? Another good question.
```
qsmmorgan: I would've preferred making them reference types, but we've talked about that before. It works either way and it's not a big deal.

philium: I started down the road of making them reference types on macOS, but it has some significant downsides. No synthesized conformance to `Equatable`, `Hashable`, `Decodable`, etc.

qsmmorgan: The `Decodable` conformance is indeed a drawback. I usually use `===` for `Equatable` classes or inherit from `NSObject`.

philium: I think I've used `===` for some classes in the past as well. But rereading the `Equatable` documentation just now, it says that's not what we're supposed to do. Fascinating.

qsmmorgan: This doc? https://developer.apple.com/documentation/swift/equatable

I don't see anything there that discourages it.

arcgis-runtime-samples-macos/Content Display Logic/Controllers/SampleListViewController.swift
```diff
@@ -16,80 +16,184 @@
...
-    /// - Returns: The node's parent or `nil` if the node does not have a parent.
-    func parent(for node: Node) -> Node? {
+    /// - Parameter sample: The sample to select.
+    func select(_ sample: Sample) {
+        guard isViewLoaded else { return }
+        guard let boxedSample = boxedSamples[sample] else { return }
+
+        let row = outlineView.row(forItem: boxedSample)
+        // Is the sample's category expanded?
+        if row != -1 {
```
nixta: This hurts my eyes in the days of Swiftiness. But I suppose it actually is the right way to do it. Surprised Apple haven't at least defined a constant for it.

qsmmorgan: That's life with `AppKit`!

arcgis-runtime-samples-macos/Content Display Logic/Controllers/SampleListViewController.swift
```diff
@@ -16,80 +16,184 @@
...
+        guard isViewLoaded else { return false }
+        guard let boxedCategory = boxedCategories[category] else { return false }
+        return outlineView.isItemExpanded(boxedCategory)
+    }
+    
+    @IBOutlet private weak var outlineView: NSOutlineView!
+    
+    private func categoriesDidChange() {
+        // Box categories and samples.
+        var boxedCategories = [Category: Box<Category>]()
```
nixta: It's a little confusing while learning the code that `Box` is used here, before it's been declared, even though it's an incredibly simple class.

arcgis-runtime-samples-macos/Content Display Logic/Controllers/SampleListViewController.swift
```diff
@@ -16,80 +16,184 @@
...
+        var boxedSamples = [Sample: Box<Sample>]()
+        for category in categories {
+            boxedCategories[category] = Box(value: category)
+            for sample in category.samples {
+                boxedSamples[sample] = Box(value: sample)
+            }
+        }
+        self.boxedCategories = boxedCategories
```
nixta: It's also quite confusing that `SampleListViewController.boxedCategories` hasn't been defined yet, but at the same time we're overriding with local function scope variables of the same name.

arcgis-runtime-samples-macos/Content Display Logic/Controllers/SampleListViewController.swift
qsmmorgan: Okay! I can see how it doesn't encourage use of `===`, but it also isn't explicit in discouraging it. I'm also thinking of how in practice, every Swift `class` I've seen compares via reference. Interesting how the doc isn't aligned with that behavior, but in this case I wouldn't take the doc as the gold standard. So many algorithms and APIs would break if comparing via value.

arcgis-runtime-samples-macos/Content Display Logic/Controllers/SampleListViewController.swift
```diff
@@ -16,80 +16,184 @@
...
+    // address, two boxed instances of the same value would never be equal.
+    // By boxing ourselves, we ensure that item comparison actually works.
+    //
+    // Why not make Category and Sample reference types? Another good question.
+    // There are a number of reasons, but one reason is because they don't need
+    // to be reference types anywhere else. Using value types is only
+    // problematic for this class, so I've chosen to solve the problem inside of
+    // this class. -Philip
+    
+    private class Box<T> {
```
nixta: I guess I would prefer that this class were named a little more descriptively. Perhaps `OutlineViewItem` so that where it's used the code seems a little more understandable? Not 100% sure if that's the right name for it, but `Box` seems far too generic.

Actually, if I rename this to `OutlineViewItem`, and rename variables beginning with `boxed` to beginning with `outlineView`, a lot of this code becomes a lot clearer to me. I think that's what was bothering me.

arcgis-runtime-samples-macos/Content Display Logic/Controllers/SampleListViewController.swift
```diff
@@ -16,80 +16,184 @@
...
         }
     }
     
     func outlineView(_ outlineView: NSOutlineView, isItemExpandable item: Any) -> Bool {
-        let node = item as! Node
-        return !node.childNodes.isEmpty
+        return item is Box<Category>
     }
     
     func outlineView(_ outlineView: NSOutlineView, child index: Int, ofItem item: Any?) -> Any {
```
nixta: Could likewise shift the unboxing into the `if let`:
``` Swift
func outlineView(_ outlineView: NSOutlineView, child index: Int, ofItem item: Any?) -> Any {
    if let category = (item as? OutlineViewItem<Category>)?.value {
        return outlineViewSamples[category.samples[index]]!
    } else {
        return outlineViewCategories[categories[index]]!
    }
}
```

I think it provides a little less benefit compared with the previous example, but still a bit cleaner.

nixta: Hmmm. I must have typed my review comment but not posted it about the `outlineView(numberOfChildrenOfItem:)` func above… I meant to suggest this this was a bit clearer:

``` Swift
func outlineView(_ outlineView: NSOutlineView, numberOfChildrenOfItem item: Any?) -> Int {
    if let category = (item as? OutlineViewItem<Category>)?.value {
        return category.samples.count
    } else {
        return categories.count
    }
}
```

PR #177, opened 2018-09-19
Fixes an issue where sample source filenames were not decoded

qsmmorgan:


qsmmorgan:
No problem! I couldn't figure out why the source wasn't loading on my new sample, then I realized it wasn't loading on any sample. The perils of data import!

philium:
It's my fault. I should have copied the XML from the macOS plist instead of modifying the XML from the iOS tests.

PR #178, opened 2018-09-19
Adds the Display KML sample

qsmmorgan:
I found that this sample would often cause crashes, but the cause appeared to be in the runtime core.


arcgis-runtime-samples-macos/Layers/Display KML/README.md
```diff
@@ -0,0 +1,28 @@
...
+## Relevant API
+
+* `AGSKMLDataset(url: URL)`
+* `AGSKMLDataset(name: String)`
+* `AGSKMLLayer(kmlDataset: AGSKMLDataset)`
+* `AGSKMLLayer(item: AGSItem)`
+
+## About the data
+
+The sample displays three different KML file, each from a different source:
```
pgruenler: `file` should be `files`

arcgis-runtime-samples-macos/Layers/Display KML/DisplayKMLViewController.swift
```diff
@@ -0,0 +1,97 @@
+//
```
pgruenler: Samples copyright notice

arcgis-runtime-samples-macos/Layers/Display KML/DisplayKMLViewController.swift
```diff
@@ -0,0 +1,97 @@
...
+        kmlLayer.load {[weak self] (error) in
+            
+            // If another layer started loading before this one finished, don't proceed
+            guard self?.loadingLayer == kmlLayer else{
+                return
+            }
+                
+            NSApp.hideProgressIndicator()
+            
+            guard let self = self else{
```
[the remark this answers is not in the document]
qsmmorgan: The progress indicator has to be hidden even if `self` is `nil`, since the user could navigate away from the sample before `load` returns and leave the indicator spinning indefinitely. This was one reason I moved the progress indicator functions to `NSApplication` from `NSWindow`.

I was able to clean up the closure a bit more. Thanks for spotting that!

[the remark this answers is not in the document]
philium: > the user could navigate away from the sample before load returns and leave the indicator spinning indefinitely.

That's true. But one of the unfortunate realities of using a shared indicator is that nobody owns it. Who's to say another sample hasn't been opened and is now loading something? That's probably unlikely, but certainly possible. The user may have a horrid network connection and the layer doesn't load right away. They decide to open a different sample that also loads a layer. The first sample's layer finally finishes loading (or fails), turning off the indicator for the second sample.

A better solution would probably be to have the main view controller hide the progress indicator whenever the selection in the sidebar changes.

> Thanks for spotting that!

You're welcome.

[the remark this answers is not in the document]
qsmmorgan: Hmm you're right, that would be a better solution! Though I'm thinking more and more that having a global progress indicator isn't the best idea.

qsmmorgan:
@pgruenler I made the changes you requested, when you get a chance to re-review

qsmmorgan:
@saratchandrakarumuri This sample is ready for testing. Note that the URL source often crashes the app, but this is an issue in the core.

arcgis-runtime-samples-macos/Info.plist
```diff
@@ -28,6 +28,13 @@
 	<dict>
 		<key>NSExceptionDomains</key>
 		<dict>
+			<key>noaa.gov</key>
```
philium: Did you manually update this file? That's fine because I know it can be quicker than using Xcode's plist editor. However, Xcode likes to keep dictionaries sorted in alphabetical order by key. That means that the next person to edit this file will be prompted by Xcode to move this key value pair.

Could you please update this file to match Xcode's style? One way to make sure your changes use the correct style is to just open the Info settings and modify any entry. Once Xcode has regenerated the file, manually revert your modification. Xcode will automatically re-sort the keys when it regenerates the file.

qsmmorgan: I used Xcode's editor but it didn't sort the keys or give a warning. Hmm! I'll fix it, thanks.

qsmmorgan:
The core has been updated with a fix for the crash.

PR #179, opened 2018-09-19
Enhances app-wide search

qsmmorgan:
Including:
- Instant results updating
- Searching of sample titles and descriptions
- Deselection of all outline view items upon display of search results
- Multi-word search support
- Search results ranking by relevancy
- No duplicate results due to the featured category

Also removes the featured category.


arcgis-runtime-samples-macos/AppDelegate.swift
```diff
@@ -47,7 +47,10 @@ class AppDelegate: NSObject {
 extension AppDelegate: NSApplicationDelegate {
     func applicationDidFinishLaunching(_ notification: Notification) {
         // Decode and populate Categories.
-        mainViewController.categories = decodeCategories(at: contentPlistURL)
+        let categories = decodeCategories(at: contentPlistURL)
+        mainViewController.categories = categories
+        let allSamples = categories.flatMap({ $0.samples })
+        mainWindowController.loadSearchEngine(samples: allSamples)
```
philium: I'm curious. Why couldn't the `mainWindowController` create the search engine itself when `categories` is assigned?

arcgis-runtime-samples-macos/Content Display Logic/Controllers/MainViewController.swift
```diff
@@ -38,16 +39,30 @@ class MainViewController: NSSplitViewController {
     /// Shows the given category in the right split view item.
     ///
     /// - Parameter category: A category.
-    func show(_ category: Category) {
-        let sampleCollectionViewController = SampleCollectionViewController(samples: category.samples)
+    func showCollection(for category: Category) {
```
philium: Could you explain why you renamed this method? I'm not sure its necessary to encode how the category will be shown, namely with a collection view, in the method signature.

arcgis-runtime-samples-macos/Content Display Logic/Controllers/MainViewController.swift
```diff
@@ -38,16 +39,30 @@ class MainViewController: NSSplitViewController {
...
+        showCollection(for: category.samples)
+    }
+    
+    /// Shows a collection view with the samples in the right split view item.
+    ///
+    /// - Parameter samples: The samples to display.
+    func showCollection(for samples: [Sample]) {
```
philium: What's the advantage to having this method? Both this and the previous method take different parameters but do exactly the same thing. I'm not sure we need both.

arcgis-runtime-samples-macos/Content Display Logic/Controllers/MainViewController.swift
```diff
@@ -38,16 +39,30 @@ class MainViewController: NSSplitViewController {
...
+        let sampleCollectionViewController = SampleCollectionViewController(samples: samples)
         sampleCollectionViewController.delegate = self
         showDetailViewController(sampleCollectionViewController)
     }
     
+    
+    /// Shows a collection view with all the samples in the app.
+    func showCollectionForAllSamples() {
```
philium: Is this a behavior change? I don't remember ever displaying all samples to the user.

arcgis-runtime-samples-macos/Content Display Logic/Controllers/MainViewController.swift
```diff
@@ -38,16 +39,30 @@ class MainViewController: NSSplitViewController {
...
+        let allSamples = categories.flatMap({ $0.samples })
+        showCollection(for: allSamples)
+    }
+    
     /// Shows the given sample in the right split view item.
     ///
     /// - Parameter sample: A sample.
-    func show(_ sample: Sample) {
+    func show(sample: Sample) {
```
philium: This function name doesn't conform to the [Swift API Design Guidelines](
https://swift.org/documentation/api-design-guidelines/#argument-labels). Although to be fair, neither did the name I came up with. Because the first argument forms part of a grammatical phrase, it shouldn't have a label and the method's name should be `showSample(_ sample: Sample)`.

arcgis-runtime-samples-macos/Content Display Logic/Controllers/WindowController.swift
```diff
@@ -60,100 +50,32 @@ class WindowController: NSWindowController, NSSearchFieldDelegate, NSWindowDeleg
...
+            let matchingSamples = searchEngine.sortedSamples(matching: query)
+            mainViewController.showCollection(for: matchingSamples)
         }
-    }
-    
-    func hideSuggestionsWindow() {
-        if let window = self.suggestionsWindowController?.window , window.isVisible {
-            window.orderOut(self)
+        else{
+            mainViewController.showCollectionForAllSamples()
```
philium: I'm not sure we should show all samples in all categories when the search is empty. It looks like you've unselected whatever was selected in the sidebar, so perhaps we should show no samples since nothing is selected. We could show a label in the middle with a message like "No Category or Sample Selected", similar to when no message is selected in Mail.

arcgis-runtime-samples-macos/Content Display Logic/Model/SampleSearchEngine.swift
```diff
@@ -0,0 +1,193 @@
+// Copyright 2016 Esri.
```
philium: I didn't look to carefully at this file. I'm assuming it is now essentially the same as the iOS implementation?

qsmmorgan:
I think the search suggestions are unnecessary now that the results update instantly as you type. Plus removing them get saves a lot of custom code and UI. I had previously removed them from iOS search.

arcgis-runtime-samples-macos/Content Display Logic/Model/SampleSearchEngine.swift
qsmmorgan: It is exactly the same aside from the import statement.

arcgis-runtime-samples-macos/Content Display Logic/Controllers/MainViewController.swift
qsmmorgan: I'm all for having a single method if it's this one, since it has a broader use.

arcgis-runtime-samples-macos/Content Display Logic/Controllers/MainViewController.swift
philium: That's the advantage of having `showCategory(_:)`. I'd really rather not have a `showSamples(_:)` method for that very reason. As I demonstrated, it is simple enough to create a category for search results.

arcgis-runtime-samples-macos/Content Display Logic/Controllers/WindowController.swift
qsmmorgan: The idea behind this is that when the search field is empty, the user is seeing the entirety of the samples that will be filtered as they type. It only show after the user types something in the search field and then deletes their input.

It doesn't matter to me either way whether we show everything or an empty collection view. It's already implemented this way, but I'm fine changing it if you prefer.

arcgis-runtime-samples-macos/Content Display Logic/Controllers/MainViewController.swift
qsmmorgan: I don't like the idea of creating ad hoc categories outside of the data import when all that's passed to the collection view anyway is the samples themselves. It's an unnecessary wrapping and unwrapping.

arcgis-runtime-samples-macos/AppDelegate.swift
qsmmorgan: The names make me bug-eyed but `categories` is assigned to the `mainViewController` and `loadSearchEngine` is called on `mainWindowController`. The view controller doesn't directly have a reference to the window controller so this seemed like the cleanest solution.

arcgis-runtime-samples-macos/Content Display Logic/Controllers/MainViewController.swift
qsmmorgan: I'll defer to you though. It doesn't really matter to me.

arcgis-runtime-samples-macos/AppDelegate.swift
philium: I see what you mean. Wow, I totally read `mainViewController` and `mainWindowController` as the same thing. Maybe we should hand the categories off to the main WINDOW controller and it could be responsible for passing things off to its children, namely main VIEW controller? You don't have to do that now. Just an idea for the future.

arcgis-runtime-samples-macos/Content Display Logic/Controllers/MainViewController.swift
philium: I think I would prefer to have the following two methods:

```swift
func showCategory(_ category: Category)
func showSample(_ sample: Sample)
```

Yes, since the category is only needed to get at the samples, the first method could be `func showSamples<S: Sequence>(_ samples: S) where S.Element == Sample`. But as you've pointed out, having `showSample(_:)` and `showSamples(_:_` is a little confusing. Also, the two methods correspond to the common use case, which is users selecting either a category or sample in the sidebar. Displaying search results is really a special case and could be thought of as a "Search Results" category.

arcgis-runtime-samples-macos/Content Display Logic/Controllers/WindowController.swift
qsmmorgan: Any strong preference on this?

qsmmorgan: Another option is to remember what was selected before searching and to re-display that when the search field is cleared.

philium: I suppose I don't really have a preference. My only concern was that there's no other way to get that behavior other than to start a search. I don't want to inadvertently create a feature with a weird method of invoking it. Perhaps we should allow deselection in the sidebar and display all samples when nothing is selected?

qsmmorgan: Oh that's a good idea. I hadn't considered that.

qsmmorgan: Okay, I updated the behavior to show all samples when there is no selection.

qsmmorgan:
@philium I believe I made all the changes you asked for, so this should be ready for re-review.

PR #185, opened 2018-09-24
Adds the list KML contents sample

qsmmorgan:


arcgis-runtime-samples-macos/Layers/List KML contents/ListKMLContentsViewController.swift
```diff
@@ -0,0 +1,158 @@
...
+        if let networkLink = node as? AGSKMLNetworkLink {
+            children.append(contentsOf: networkLink.childNodes)
+        }
+        return children
+    }
+
+    /// Sets the viewpoint of the scene to the extent of the node, if it has one
+    func setSceneViewpoint(on node: AGSKMLNode){
+        if let extent = node.extent,
+            // some nodes do not include a geography so check that the extent isn't empty
```
pgruenler: I think "geometry" rather than "geography"

qsmmorgan: Thanks. Changed!

qsmmorgan:
@saratchandrakarumuri I put the KMZ file in the sample folder because I didn't know there was a prevailing pattern. I moved it in my latest commit. The other issues may be a consequence of the data since my implementation follows the spec.

nCastle1:
@qsmmorgan @saratchandrakarumuri 

* The science of where screen overlay should show on the bottom left and looks like the Esri logo; if you're not seeing it, that may be a core issue. I don't see the attribution bar at the bottom either, so maybe its just cropped out. If there is UI overlapping the sceneview, consider setting view insets to account for that. 
* The rendering on the extrusion for building E looks glitchy. There might be a core bug. 
* Re: zooming to extent, there is a [proposed change to the sample design](https://devtopia.esri.com/runtime/common-samples/pull/915) that should change the behavior. Note that the disappearance of the placemark happens with that additional behavior on .NET, so this may indicate a core bug.

arcgis-runtime-samples-macos/Layers/List KML contents/ListKMLContentsViewController.swift
```diff
@@ -0,0 +1,158 @@
...
+        }
+    }
+    
+    /// Returns all the child nodes of this node
+    func childNodes(of node: AGSKMLNode) -> [AGSKMLNode] {
+        var children = [AGSKMLNode]()
+        if let container = node as? AGSKMLContainer {
+            children.append(contentsOf: container.childNodes)
+        }
+        if let networkLink = node as? AGSKMLNetworkLink {
```
philium: `node` can be a subclass of `AGSKMLContainer` or of `AGSKMLNetworkLink`, but not both. Please either make this an `else if` or just use a `switch` statement:
```swift
switch node {
case let container as AGSKMLContainer:
    return container.childNodes
case let networkLink as AGSKMLNetworkLink:
    return networkLink.childNodes
default:
    return []
}
```

arcgis-runtime-samples-macos/Layers/List KML contents/ListKMLContentsViewController.swift
```diff
@@ -0,0 +1,158 @@
...
+            children.append(contentsOf: networkLink.childNodes)
+        }
+        return children
+    }
+
+    /// Sets the viewpoint of the scene to the extent of the node, if it has one
+    func setSceneViewpoint(on node: AGSKMLNode){
+        if let extent = node.extent,
```
philium: Consider replacing with a `guard` statement.

arcgis-runtime-samples-macos/Layers/List KML contents/ListKMLContentsViewController.swift
```diff
@@ -0,0 +1,158 @@
...
+    }
+    
+    func outlineView(_ outlineView: NSOutlineView, child index: Int, ofItem item: Any?) -> Any {
+        if let node = item as? AGSKMLNode {
+            return childNodes(of: node)[index]
+        }
+        else if let kmlDataset = kmlDataset {
+            return kmlDataset.rootNodes[index]
+        }
+        return NSNull()
```
philium: This is essentially a never-should-happen case, right? If so, I recommend just using `else { fatalError() }`. That way if the won't ever happen does happen as a result of a development change, it will be caught immediately.

arcgis-runtime-samples-macos/Layers/List KML contents/ListKMLContentsViewController.swift
```diff
@@ -0,0 +1,158 @@
...
+           
+            let viewpoint = AGSViewpoint(targetExtent: extent)
+            sceneView.setViewpoint(viewpoint)
+        }
+    }
+    
+}
+
+// Boilerplate datasource code to display the KML node hierarchy
+extension ListKMLContentsViewController: NSOutlineViewDataSource {
```
philium: You might consider making a separate type for the data source:
```swift
class OutlineViewDataSource: NSOutlineViewDataSource {
    let kmlDataset: AGSKMLDataset
    
    init(kmlDataset: AGSKMLDataset) {
        self.kmlDataset = kmlDataset
    }
    
    …
}
```
And only creating it and assigning it to the outline view once the data set has loaded. That way the data source doesn't have to deal with the case where the data set is `nil`.

arcgis-runtime-samples-macos/Layers/List KML contents/ListKMLContents.storyboard
```diff
@@ -0,0 +1,164 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
philium: Xcode is giving me a Misplaced Views warning for this file.
screen shot 2018-10-01 at 11 47 16 am

arcgis-runtime-samples-macos/Layers/List KML contents/ListKMLContents.storyboard
```diff
@@ -0,0 +1,164 @@
...
+                        <autoresizingMask key="autoresizingMask"/>
+                        <subviews>
+                            <splitView arrangesAllSubviews="NO" dividerStyle="thin" vertical="YES" translatesAutoresizingMaskIntoConstraints="NO" id="uZe-TP-J5V">
+                                <rect key="frame" x="0.0" y="0.0" width="501" height="329"/>
+                                <subviews>
+                                    <customView misplaced="YES" id="zp0-CP-Mun">
+                                        <rect key="frame" x="0.0" y="0.0" width="153" height="329"/>
+                                        <autoresizingMask key="autoresizingMask" flexibleMaxX="YES" flexibleMinY="YES"/>
+                                        <subviews>
+                                            <customView wantsLayer="YES" translatesAutoresizingMaskIntoConstraints="NO" id="mY0-P9-stY">
```
philium: I'm not sure this view is needed. Outline views by their nature afford selecting. so I don't think telling the user to make a selection adds any benefit.

arcgis-runtime-samples-macos/Layers/List KML contents/ListKMLContentsViewController.swift
qsmmorgan: I try never to use `fatalError` unless there's a reason the whole app cannot continue. If that case ever did happen, I'd rather have a blank outline view in one sample than crash entirely.

arcgis-runtime-samples-macos/Layers/List KML contents/ListKMLContentsViewController.swift
qsmmorgan: Replaced in the latest commit.

arcgis-runtime-samples-macos/Layers/List KML contents/ListKMLContentsViewController.swift
qsmmorgan: Oh okay, makes sense.

arcgis-runtime-samples-macos/Layers/List KML contents/ListKMLContentsViewController.swift
qsmmorgan: While that does have a small advantage, I'm not sure adding another type and retaining another object is worth that.

arcgis-runtime-samples-macos/Layers/List KML contents/ListKMLContents.storyboard
qsmmorgan: Hmm not seeing this on mine

arcgis-runtime-samples-macos/Layers/List KML contents/ListKMLContents.storyboard
qsmmorgan: You're right!

arcgis-runtime-samples-macos/Layers/List KML contents/ListKMLContentsViewController.swift
philium: I hear what you're saying. And it is rather scary to see `fatalError()` in production code. But just because it's there doesn't mean it will be be executed. I'd rather the app crash on us (as developers) to tell us that we've done something wrong rather than have a silent failure go overlooked. Do you know how `NSOutlineView` handles `NSNull`? It might just fail anyway.

philium: My experience with `NSOutlineView` is fairly limited, but based on what I know, this is somewhat analogous to `cellForRowAtIndexPath`. The data source has already said that there's a data model object at the given index. That object is either a top level object or a child object. Returning `NSNull` would be the equivalent of returning `nil` (if the API would let you) when a cell is requested. That's crazy! The data source is essentially saying that there is and there is not something there.

From the documentation for `outlineView(_:child:ofItem:):

> The child item at index of item. If item is nil, returns the appropriate child item of the root object.

There is no `nil` option. That's a big advantage I see to using a separate object initialized with a non-`nil` data set. The data set would never be `nil`, preventing that scenario altogether:
```swift
if let node = item as? AGSKMLNode {
    return childNodes(of: node)[index]
} else {
    return dataSet.rootNodes[index]
}
```

arcgis-runtime-samples-macos/Layers/List KML contents/ListKMLContents.storyboard
philium: I'm not seeing it anymore either. I think it might have been resolved by this commit: https://github.com/Esri/arcgis-runtime-samples-macos/pull/185/commits/f8a4c6381b14f9dc0f17820d4811bb2df40f5678

arcgis-runtime-samples-macos/Layers/List KML contents/ListKMLContentsViewController.swift
qsmmorgan: I changed it so the data source is set after the data is loaded, so that `kmlDataset` can never be `nil` and the data source functions never need to account for a `nil` case.

qsmmorgan:
@saratchandrakarumuri Ready for testing again.

qsmmorgan:
@saratchandrakarumuri Thanks!

PR #186, opened 2018-09-25
Adds the Display KML Network Links sample

qsmmorgan:


arcgis-runtime-samples-macos/Layers/Display KML network links/DisplayKMLNetworkLinksViewController.swift
```diff
@@ -0,0 +1,56 @@
...
+        // set the initial text
+        textView.string = "Network Link Messages:\n\n"
+        
+        /// The URL of a KML file with network links
+        let datasetURL = URL(string: "https://www.arcgis.com/sharing/rest/content/items/600748d4464442288f6db8a4ba27dc95/data")!
+        /// The KML dataset with network links
+        let kmlDataset = AGSKMLDataset(url: datasetURL)
+        
+        // register to receive the network link messages
+        kmlDataset.networkLinkMessageHandler = {[weak self] (_, message) in
```
pgruenler: Is there anything (vaguely)interesting that you could do with the network link rather than leave an underscore in the arguments?  It might be nice to show that it is there.

qsmmorgan: I can't think of anything to do with the network link. It isn't used in the sample spec. I could just change the underscore to `kmlNetworkLink` and leave it unused? That way it's more obvious to users that it's available. Merging as-is.

PR #188, opened 2018-09-25
Uses visual effect view for collection view header

philium:
It turns out that `NSColor.headerColor` is deprecated. It also turns out that using a visual effect view makes the header look right in both light and dark modes.


qsmmorgan:
@philium I don't think it opened the issue navigator automatically, so I just happened to see the warning in passing.

PR #195, opened 2018-09-26
Viewshed (GeoElement): Uses visual effect view for instructions

philium:
This allows it to adapt to the current appearance.

### DIFF
--- a/arcgis-runtime-samples-macos/Analysis/Viewshed (GeoElement)/ViewshedGeoElement.storyboard
+++ b/arcgis-runtime-samples-macos/Analysis/Viewshed (GeoElement)/ViewshedGeoElement.storyboard
@@ -1,10 +1,9 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<document type="com.apple.InterfaceBuilder3.Cocoa.Storyboard.XIB" version="3.0" toolsVersion="13196" targetRuntime="MacOSX.Cocoa" propertyAccessControl="none" useAutolayout="YES" initialViewController="lxB-4W-Frk">
+<document type="com.apple.InterfaceBuilder3.Cocoa.Storyboard.XIB" version="3.0" toolsVersion="14313.18" targetRuntime="MacOSX.Cocoa" propertyAccessControl="none" useAutolayout="YES" initialViewController="lxB-4W-Frk">
     <dependencies>
         <deployment identifier="macosx"/>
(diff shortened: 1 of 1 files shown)


philium:
> It would be easier

@qsmmorgan Would you mind elaborating? I'm really curious to hear your perspective.

I've found that having several small PRs as opposed to one large PRs has its benefits. It allows changes to get in sooner as requested changes only hold up changes in those respective PRs rather than all changes. It also allows for easier reversion of a particular change by simply reverting that change's PR. I've also found that in general, larger PRs tend to hang around longer than smaller ones. A lot longer than they would have if those same changes were made in smaller PRs. 

As a reviewer, I'm much more comfortable putting my stamp of approval on a small number of changes than on a bunch of changes. The more changes there are, the less confident I am that I didn't miss anything.

philium:
@qsmmorgan I suppose I could group changes by category. That would make the PRs somewhat larger. Would you find that easier?

qsmmorgan:
Your points make sense, though personally, I have no problem with large PRs as long as all the changes are related. I think there's a cost of time and mental energy to creating and reviewing every PR no matter then length, so fewer PRs can sometimes be more efficient. Then again, I haven't had any of my PRs sit around for a year!

I think there are tradeoffs either way. I don't have a problem continuing with PRs of this size, I just wanted you to feel free to group some of them if you wanted.

mhdostal:
Are these the dark mode changes?
```
<visualEffectView wantsLayer="YES" appearanceType="inheritedVibrantLight" blendingMode="withinWindow" material="underWindowBackground" state="followsWindowActiveState" translatesAutoresizingMaskIntoConstraints="NO" id="lrg-R8-ISS">
```

mhdostal:
What were the `textField` size/location changes for?

philium:
> Are these the dark mode changes?

Yes. Previously the text field was embedded in a view with a hardcoded background color that made the text almost unreadable in dark mode. Embedding the text field in a visual effect view allows using a material, which adapts to the current mode, as the background color.

> What were the textField size/location changes for?

The horizontal position changed because of the way I positioned the text field in its superview. Previously the leading and trailing edges were constrained to that of its superview. Now, it is just horizontally centered in its superview and its size is determined by its intrinsic content size.

The vertical position changed because I adjusted the space between the text field and its superview. Why? Because eight is half of the default spacing and I had just worked on another sample that used eight for the spacing. That and I didn't notice the previous spacing was six until I was reviewing my changes.

PR #208, opened 2018-10-03
Identify KML Features sample

philium:
Adds the Identify KML Features sample.

Corresponding iOS PR: https://github.com/Esri/arcgis-runtime-samples-ios/pull/498


arcgis-runtime-samples-macos/Layers/Identify KML features/IdentifyKMLFeaturesViewController.swift
```diff
@@ -0,0 +1,79 @@
...
+
+extension IdentifyKMLFeaturesViewController: AGSGeoViewTouchDelegate {
+    func geoView(_ geoView: AGSGeoView, didTapAtScreenPoint screenPoint: CGPoint, mapPoint: AGSPoint) {
+        geoView.callout.dismiss()
+        geoView.identifyLayer(forecastLayer, screenPoint: screenPoint, tolerance: 15, returnPopupsOnly: false) { [weak self] (result) in
+            if let error = result.error {
+                print("Error identifying layer: \(error)")
+            } else if let placemarkIndex = result.geoElements.firstIndex(where: { $0 is AGSKMLPlacemark }) {
+                let placemark = result.geoElements[placemarkIndex] as! AGSKMLPlacemark
+                self?.showCallout(for: placemark, at: mapPoint)
```
qsmmorgan: Might consider unwrapping `self` earlier to avoid unnecessary work if it's `nil`.

arcgis-runtime-samples-macos/Layers/Identify KML features/IdentifyKMLFeaturesViewController.swift
```diff
@@ -0,0 +1,79 @@
...
+        let map = AGSMap(basemap: .darkGrayCanvasVector())
+        map.operationalLayers.add(forecastLayer)
+        
+        let center = AGSPoint(x: -48_885, y: 1_718_235, spatialReference: AGSSpatialReference(wkid: 5070))
+        map.initialViewpoint = AGSViewpoint(center: center, scale: 50_000_000)
+        
+        mapView.map = map
+    }
+    
+    func showCallout(for placemark: AGSKMLPlacemark, at point: AGSPoint) {
```
qsmmorgan: Could be a private method.

arcgis-runtime-samples-macos/Layers/Identify KML features/README.md
```diff
@@ -0,0 +1,35 @@
+# Identify KML features
+
+Show a balloon popup with formatted content for tapped KML features. KML supports defining HTML for popups.
```
qsmmorgan: Replace "tapped" to indicate the mouse paradigm.

arcgis-runtime-samples-macos/Layers/Identify KML features/README.md
```diff
@@ -0,0 +1,35 @@
...
+
+![](image1.png)
+
+## How to use the sample
+
+Click or tap a feature to identify it. Feature information will be displayed in a callout.
```
qsmmorgan: Ditto.

arcgis-runtime-samples-macos/Layers/Identify KML features/README.md
```diff
@@ -0,0 +1,35 @@
...
+
+## How to use the sample
+
+Click or tap a feature to identify it. Feature information will be displayed in a callout. 
+
+Note: the KML layer used in this sample contains a screen overlay. The screen overlay contains a legend and the logos for NOAA and the NWS. You can't identify the screen overlay.
+
+## How it works
+
+1. The identify process starts when the user taps the map.
```
qsmmorgan: Ditto.

arcgis-runtime-samples-macos/Layers/Identify KML features/IdentifyKMLFeaturesViewController.swift
philium: To be honest, I've consciously not specified access controls in most (all?) of the samples I've worked on. It is kind of like localization. Or accessibility. Is it the right thing to do? Yes! Does it make a difference in sample code? Not really. What are your thoughts?

arcgis-runtime-samples-macos/Layers/Identify KML features/README.md
philium: The reason I kept it as "tapped" is that the touch delegate method has "tap" in the name. That being said, "click" is the right verb to use. I'll go ahead and change it.

arcgis-runtime-samples-macos/Layers/Identify KML features/IdentifyKMLFeaturesViewController.swift
qsmmorgan: Haha I think localization and accessibility are far more important, but I've been trying to get better about access control because I want to work it into my regular coding habits. In the samples apps, I think access control is most important for helper type definitions since we don't want lots of types exposed between samples. But even then it's not worth worrying about. I didn't have a lot to mention in this PR so I figured I'd mention this 🤷‍♂️

arcgis-runtime-samples-macos/Layers/Identify KML features/README.md
qsmmorgan: Oh that makes sense. More dual-platform weirdness.

arcgis-runtime-samples-macos/Layers/Identify KML features/IdentifyKMLFeaturesViewController.swift
philium: 🤣 For you, I will make the method `private`.

arcgis-runtime-samples-macos/Layers/Identify KML features/IdentifyKMLFeaturesViewController.swift
philium: Good idea. It also doesn't make much sense to log an error if `self` isn't around anymore.

PR #211, opened 2018-10-04
Adapts collection view cells to dark appearance

philium:
I took the two custom blue colors and used the lighter one for light appearance and the darker one for dark appearance. Note that the cell border colors and header background colors are now the same. Previously they were different.


qsmmorgan:
I like the dark cells! To keep high contrast in the light cells, I'd keep the header text white and make the description text darker.

screen shot 2018-10-04 at 3 00 00 pm

philium:
Thank you for your feedback. I was trying to use the system-provided semantic colors as much as possible. I used primary text color for the header text and secondary text color for the description. Here's what it would look like with your proposed changes:

screen shot 2018-10-04 at 3 12 08 pm

I'm not sure I like the header and the description having opposite appearance (one being light while the other is dark). Perhaps we could use a different blue color?

philium:
@qsmmorgan Thank you. I don't feel too strongly about the description color, but I do like having it be a different color than the title in dark mode. I think I'd like to leave it the way it is for now and see what other people think. We can change it if need be.

PR #225, opened 2018-10-18
Adds the Display Layer View State sample

philium:


arcgis-runtime-samples-macos/Maps/Display layer view state/DisplayLayerViewStateViewController.swift
```diff
@@ -0,0 +1,78 @@
...
+        map.operationalLayers.addObjects(from: [tiledLayer, imageLayer, featurelayer])
+        
+        return map
+    }
+    
+    override func viewDidLoad() {
+        super.viewDidLoad()
+        
+        mapView.map = makeMap()
+        mapView.setViewpoint(AGSViewpoint(center: AGSPoint(x: -11e6, y: 45e5, spatialReference: .webMercator()), scale: 5e7))
```
nixta: Can I ask why you're using `-11e6` and `45e5` here? I think it's especially weird with the scale. I don't think anyone says "my map scale is 1:5e7".

Especially when above, in `makeMap()` you're using `40000000` and `2000000` for `minScale` and `maxScale`.

I would prefer the coordinates and scale be written out in full.

philium: > Can I ask why you're using -11e6 and 45e5 here?

I just copied what the iOS implementation had.

philium:
@qsmmorgan I didn't add `LayerViewStatusPanel.swift` to the source file list because it doesn't contain any runtime-specific code. See the distance measurement analysis sample for a similar approach.

qsmmorgan:
Up to you, but I'd have included it.

PR #230, opened 2018-10-23
Enables linting via SwiftLint

philium:
Also adds configuration file that disables most of the default rules. These will be enabled individually to aid in discussion and review of resulting changes.


philium:
@qsmmorgan We have approval to move forward with this, so please re-approve when you get a chance.

PR #231, opened 2018-10-23
Fixes errors in unit testing target

philium:


qsmmorgan:
Also, my bad for not running the tests when changing the name.

PR #233, opened 2018-10-26
Replaces KML node class names with user-friendly labels in List KML Contents

qsmmorgan:


arcgis-runtime-samples-macos/Layers/List KML contents/ListKMLContentsViewController.swift
```diff
@@ -96,6 +96,25 @@ class ListKMLContentsViewController: NSViewController {
         return []
     }
     
+    /// Returns a label `String` for the type of the `node` based on the `class`.
+    /// For example, a `node` with `class` `AGSKMLScreenOverlay` returns "Screen Overlay".
+    private func typeLabel(for node: AGSKMLNode) -> String {
+        var classNameString = "\(type(of: node))"
```
philium: Parsing the type name is fun and all, but I'm not sure it is something we want to model. This would not be a good idea in almost any production app because the label would not be localized. Since the list of possible nodes is finite, small, and known at compile time, it might be best to just hardcode the titles.

qsmmorgan: True. I can change it.

PR #234, opened 2018-10-31
Adds the Open Mobile Map (Map Package) sample

qsmmorgan:


arcgis-runtime-samples-macos/Maps/Open mobile map (map package)/OpenMobileMapViewController.swift
```diff
@@ -0,0 +1,51 @@
...
+        //load map package
+        mapPackage.load { [weak self] (error: Error?) in
+            guard let self = self else {
+                return
+            }
+
+            if let error = error {
+                NSAlert(error: error).beginSheetModal(for: self.view.window!)
+            }
+            else if !mapPackage.maps.isEmpty {
```
philium: You could have performed this check and retrieved the map at the same time:
```swift
else if let map = mapPackage.maps.first {
    self.mapView.map = map
}
```
But since `AGSMapView.map` is optional, I would suggest just using an else clause:
```swift
else {
    self.mapView.map = mapPackage.maps.first
}
```
We know that the map package has a map, so in the event that the package loads without error, `mapPackage.maps.isEmpty` will never be `true`. Therefore we shouldn't plan for that.

arcgis-runtime-samples-macos/Maps/Open mobile map (map package)/OpenMobileMapViewController.swift
```diff
@@ -0,0 +1,51 @@
+//
+// Copyright 2016 Esri.
```
philium: `2016` -> `2018`

arcgis-runtime-samples-macos/Maps/Open mobile map (map package)/OpenMobileMapViewController.swift
qsmmorgan: Good call. I didn't put much thought into it since it was copied from iOS.

qsmmorgan:
@philium I made the requested changes.

qsmmorgan:
@saratchandrakarumuri Thanks!

PR #240, opened 2018-11-05
Enables the unused_closure_parameter rule

philium:


arcgis-runtime-samples-macos/Display information/Show legend/ShowLegendViewController.swift
```diff
@@ -52,7 +52,7 @@ class ShowLegendViewController: NSViewController, NSOutlineViewDataSource, NSOut
         //add feature layer to the map
         self.map.operationalLayers.add(featureLayer)
         
-        AGSLoadObjects(self.map.operationalLayers as AnyObject as! [AGSLayer]) { [weak self] (success) in
+        AGSLoadObjects(self.map.operationalLayers as! [AGSLayer]) { [weak self] (_) in
```
pgruenler: This change makes me wonder if linting is causing changes that are not quite in line with the purpose of the samples to teach the API.  By replacing `success` with `_` I feel I would learn a bit less about the API by looking at this sample.  Nothing I couldn't get other ways but I want to make sure linting doesn't cause the samples to loose their focus.

arcgis-runtime-samples-macos/Edit data/Use geodatabase transactions/UseGeodatabaseTransactionsViewController.swift
```diff
@@ -216,7 +216,7 @@ class UseGeodatabaseTransactionsViewController: NSViewController {
         presentAsSheet(progressViewController)
         
         // start the upload
-        syncGeodatabaseJob.start(statusHandler: nil) {[weak self] (results, error) in
+        syncGeodatabaseJob.start(statusHandler: nil) { [weak self] (_, error) in
```
pgruenler: Same comment as above, replace `results` with `_`.

arcgis-runtime-samples-macos/Display information/Show legend/ShowLegendViewController.swift
philium: Using `_` in place of an unused parameter is a Swift compiler optimization used throughout the samples app. I suspect that, at least when it comes to runtime API calls, we should instead be using the parameters in some way rather than leaving them unused. However, for most (if not all) of these API calls, we have examples in other samples of the same call where the parameters are used.

pgruenler: > Using _ in place of an unused parameter is a Swift compiler optimization used throughout the samples app.

It's not that I don't understand the change; it's that I'm not completely convinced of its value in this context.  Swift compiler optimizations are good but I'm just trying to make sure you have an eye on instructional effectiveness too.

qsmmorgan: I tend to agree with @pgruenler on this one. We have to be careful about making tradeoffs between performance and readability.

pgruenler: @dg0yal, do you have any thoughts on this?

philium: @pgruenler We've been waiting quite a while for a response. Since there are other places (not in this PR) where `_` is used in place of unused parameters in `ArcGIS` function calls, would you be okay with approving this PR and then creating an issue with your concerns?

dg0yal: I don't like this change either. Ideally our samples should be handling the outcomes. If we don't want to handle every single possibility that's fine, but let the reader know what the possibilities are.  Sweeping it under the rug with a _ is counter productive.

philium:
@pgruenler I've updated the PR to use the previously unused parameters.

PR #247, opened 2018-11-13
Uses AGSSpatialReference.webMercator() consistently

philium:
Also adds custom linting rule to catch inconsistent uses in the future.


nixta:
Nice!

I know this repo doesn’t use it but would it make sense to add a linting rule for 4326 and WGS84?

PR #251, opened 2018-11-13
Enables the force_try SwiftLint rule

qsmmorgan:
Also fixes two colon rule violations


arcgis-runtime-samples-macos/Route & Directions/Find route/FindRouteViewController.swift
```diff
@@ -29,10 +29,10 @@ class FindRouteViewController: NSViewController {
     var stopGraphicsOverlay = AGSGraphicsOverlay()
     var routeGraphicsOverlay = AGSGraphicsOverlay()
     
-    var stop1Geometry:AGSPoint {
+    var stop1Geometry: AGSPoint {
```
philium: Thanks for fixing my mistake.

PR #253, opened 2018-11-13
Enables the mark, multiple_closures_with_trailing_closure, and vertical_whitespace SwiftLint rules

qsmmorgan:
There are lots of changes but they are mechanical. `multiple_closures_with_trailing_closure` needed changes in only two places.


arcgis-runtime-samples-macos/Maps/Create and save a map/SaveMapViewController.swift
```diff
@@ -34,7 +34,6 @@ class SaveMapViewController: NSViewController {
     override func viewDidLoad() {
```
philium: It would be nice to just go ahead and remove this unneeded override.

PR #254, opened 2018-11-13
Enables the block_based_kvo SwiftLint rule

qsmmorgan:
Removes an empty viewDidLoad override
Replaces the numero symbol with the degree symbol in Animate 3D Symbols

### DIFF
--- a/arcgis-runtime-samples-macos/Scenes/Animate 3D symbols/Animate3DSymbolsVC.swift
+++ b/arcgis-runtime-samples-macos/Scenes/Animate 3D symbols/Animate3DSymbolsVC.swift
@@ -136,31 +136,42 @@ class Animate3DSymbolsVC: NSViewController {
         //set the camera controller on scene view
         self.sceneView.cameraController = self.orbitGeoElementCameraController
         
+    }
+    
+    override func viewWillAppear() {
+        super.viewWillAppear()
         //add observers to update the sliders
-        cameraDistanceObservation = orbitGeoElementCameraController.observe(\.cameraDistance, options: .new) {[weak self] (controller, change) in
-            DispatchQueue.main.async { [weak self] in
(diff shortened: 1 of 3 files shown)


arcgis-runtime-samples-macos/Scenes/Animate 3D symbols/Animate3DSymbolsVC.swift
```diff
@@ -133,9 +137,30 @@ class Animate3DSymbolsVC: NSViewController {
         self.sceneView.cameraController = self.orbitGeoElementCameraController
         
         //add observers to update the sliders
-        self.orbitGeoElementCameraController.addObserver(self, forKeyPath: "cameraDistance", options: .new, context: nil)
-        self.orbitGeoElementCameraController.addObserver(self, forKeyPath: "cameraHeadingOffset", options: .new, context: nil)
-        self.orbitGeoElementCameraController.addObserver(self, forKeyPath: "cameraPitchOffset", options: .new, context: nil)
+        cameraDistanceObservation = orbitGeoElementCameraController.observe(\.cameraDistance, options: .new) {[weak self] (controller, change) in
```
philium: Please do the following:
- Change `options` from `.new` to `.initial`.
- Add a space after the opening curly brace (`{`).
- Replace `change` with `_`.
- Move these calls to `observe` into `viewWillAppear()` and set the observations to `nil` in `viewDidDisappear()`.

arcgis-runtime-samples-macos/Scenes/Animate 3D symbols/Animate3DSymbolsVC.swift
```diff
@@ -133,9 +137,30 @@ class Animate3DSymbolsVC: NSViewController {
...
+            DispatchQueue.main.async { [weak self] in
```
philium: Please remove `[weak self] in`. It is not needed since `self` is already `weak` in the enclosing scope.

PR #255, opened 2018-11-13
Enables opt-in SwiftLint rules that require no or few changes

qsmmorgan:
Reference for the enabled rules: https://github.com/realm/SwiftLint/blob/master/Rules.md


arcgis-runtime-samples-macos/Content Display Logic/Controllers/SampleListViewController.swift
```diff
@@ -178,7 +178,7 @@ extension SampleListViewController: NSOutlineViewDataSource {
         case let sampleItem as SampleItem:
             return sampleItem.value.name
         default:
-            fatalError()
```
philium: Which rule prompted this change? `fatalError()` is appropriate here because it represents an impossible condition. `item` will either be a `CategoryItem` or a `SampleItem`, but there is no way for the compiler to know that. Returning `nil` makes it seem like `default` is an expected condition.

.swiftlint.yml
```diff
@@ -17,3 +17,49 @@ disabled_rules: # rule identifiers to exclude from running
...
+  - closure_spacing
+  - collection_alignment
+  - contains_over_first_not_nil
+  - convenience_type
+  - discouraged_direct_init
+  - discouraged_optional_boolean
+  - empty_string
+  - empty_xctest_method
+  - explicit_init
+  - extension_access_modifier
```
philium: I'm not crazy about this rule. While it makes sense for extensions of `ArcGIS` types to be private, there are examples (particularly in the iOS app) of internal-by-default extensions that should be internal. I don't like the idea of requiring those to explicitly declare their access as `internal`. 

It seems to me we could create a custom rule that requires extensions of `ArcGIS` types (not conforming to a protocol) to be private.

esreli:
I have no experience using SwiftLint, removing myself from this PR.

.swiftlint.yml
```diff
@@ -17,3 +17,49 @@ disabled_rules: # rule identifiers to exclude from running
...
+  - fallthrough
+  - fatal_error_message
+  - first_where
+  - function_default_parameter_at_end
+  - identical_operands
+  - joined_default_parameter
+  - legacy_random
+  - let_var_whitespace
+  - literal_expression_end_indentation
+  - lower_acl_than_parent
```
philium: I think this is now done automatically by the compiler.

.swiftlint.yml
```diff
@@ -17,3 +17,49 @@ disabled_rules: # rule identifiers to exclude from running
...
+  - modifier_order
+  - multiline_arguments
+  - multiline_function_chains
+  - multiline_parameters
+  - overridden_super_call
+  - override_in_extension
+  - prohibited_super_call
+  - quick_discouraged_call
```
philium: This (and the following two rules) don't even make sense because we don't use `Quick`.

.swiftlint.yml
```diff
@@ -17,3 +17,49 @@ disabled_rules: # rule identifiers to exclude from running
...
+  - quick_discouraged_focused_test
+  - quick_discouraged_pending_test
+  - redundant_nil_coalescing
+  - redundant_type_annotation
+  - required_enum_case
```
philium: I don't even understand this rule.

.swiftlint.yml
```diff
@@ -17,3 +17,49 @@ disabled_rules: # rule identifiers to exclude from running
...
+  - sorted_first_last
+  - static_operator
+  - switch_case_on_newline
```
philium: I'm torn over this one. I like the idea in principle, but I think there are times where it is appropriate, particularly when each case performs the same single-line action. And since there was only one place where this rule was being violated, we already follow it 99% of the time. Therefore I don't think enabling it is super helpful.

arcgis-runtime-samples-macos/Content Display Logic/Controllers/SampleListViewController.swift
qsmmorgan: `fatal_error_message` flagged the call. I didn't see any harm in replacing it instead of writing a message since the code will never be run anyway.

.swiftlint.yml
qsmmorgan: Hmm not that I saw. I implemented the same rule in the iOS implementation and it flagged a few cases where properties were `public`.

.swiftlint.yml
qsmmorgan: Okay, do you want me to remove them?

.swiftlint.yml
qsmmorgan: Seems like it only has to do with the `NetworkResponsable` protocol.

.swiftlint.yml
qsmmorgan: I don't think SwiftLint's description of this rule is great. I doesn't required explicit access declarations. It just says if you do declare access modifiers for extensions, they should be on the extension itself rather than its properties.

If you want me to remove the rule anyway that's fine.

.swiftlint.yml
qsmmorgan: Well we already follow lots of the rules and I still think they're worth enabling, but I agree there are times when newlines aren't needed. I'll remove this rule.

.swiftlint.yml
qsmmorgan: I did.

.swiftlint.yml
qsmmorgan: I removed it.

.swiftlint.yml
philium: Yes please. Thank you.

.swiftlint.yml
philium: Thanks.

.swiftlint.yml
philium: Huh. That is…surprising. Well, if a linting rule can make up for a deficiency in the compiler, then so be it!

.swiftlint.yml
philium: Thank you. That makes a lot more sense. That sounds like a fine rule.

arcgis-runtime-samples-macos/Content Display Logic/Controllers/SampleListViewController.swift
philium: I would prefer either not enabling that rule or modify this usage of `fatalError()` to include a message, something like this:

```suggestion
            fatalError("item should only ever be a CategoryItem or SampleItem")
```

qsmmorgan: Sure. Done.

qsmmorgan:
@philium All changes should be resolved now.

PR #259, opened 2018-11-13
Enables the implicit_return, operator_whitespace, and operator_usage_whitespace rules

qsmmorgan:
Also enables several others rules that did not require code changes

### DIFF
--- a/.swiftlint.yml
+++ b/.swiftlint.yml
@@ -24,6 +24,8 @@ opt_in_rules: # some rules are only opt-in
   - anyobject_protocol
   - array_init
   - attributes
+  - closure_body_length
+  - closure_end_indentation
   - closure_spacing
   - collection_alignment
   - contains_over_first_not_nil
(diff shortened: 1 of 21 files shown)


.swiftlint.yml
```diff
@@ -40,15 +42,19 @@ opt_in_rules: # some rules are only opt-in
   - first_where
   - function_default_parameter_at_end
   - identical_operands
+  - implicit_return
   - joined_default_parameter
   - legacy_random
   - let_var_whitespace
   - literal_expression_end_indentation
   - lower_acl_than_parent
+  - missing_docs
```
qsmmorgan: Turns out this rule only flags `public` declarations without doc, so it doesn't cause any warnings at the moment. We might consider an ambitious goal of adding a rule to flag all declarations missing doc, since this is a samples app after all.

.swiftlint.yml
```diff
@@ -40,15 +42,19 @@ opt_in_rules: # some rules are only opt-in
   - first_where
   - function_default_parameter_at_end
   - identical_operands
+  - implicit_return
```
philium: I'm not crazy about this one. I really like using implicit return in *single line* closures, but I think returning statements on a line of their own look really odd without an explicit `return`:

```swift
        let selectedDefinitions = Set(selectedIndexes.map { (index) -> AGSStatisticDefinition in
             statisticDefinitions[index]
         })
```

Some closures can be easily refactored to be a single line closure, line the one in `StatisticalQueryGroupAndSortViewController.swift`:

```swift
        let selectedDefinitions = Set(selectedIndexes.map { statisticDefinitions[$0] }
```

But some really make sense as a multiline closure because of length, like the one in `Model/SampleSearchEngine.swift`:

```swift
        var matchingSamples = samples.filter { (sample) -> Bool in
             return sample.name.lowercased().contains(lowercasedQuery) ||
                 sample.description.lowercased().contains(lowercasedQuery)
         }
```

qsmmorgan: Fine by me

qsmmorgan:
@philium Ready for re-review.

PR #261, opened 2018-11-14
Moves unnecessarily global constants in two samples

qsmmorgan:
This PR fixes the SwiftLint warnings by removing top-level constants entirely. In the tests file, this meant redeclaring the model objects in every test. I think this is fine since it's better to test from a clean slate each time anyway.

### DIFF
--- a/arcgis-runtime-samples-macos/Analysis/Line of sight (geoelement)/LineOfSightGeoElementViewController.swift
+++ b/arcgis-runtime-samples-macos/Analysis/Line of sight (geoelement)/LineOfSightGeoElementViewController.swift
@@ -15,9 +15,6 @@
 import Cocoa
 import ArcGIS
 
-private let observerZMin = 20.0
-private let observerZMax = 1500.0
-
 class LineOfSightGeoElementViewController: NSViewController {
     
     @IBOutlet var sceneView: AGSSceneView!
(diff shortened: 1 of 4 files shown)


philium:
I am not in favor of enabling this rule. Requiring global constants to be prefixed with `k` is a concept foreign to Swift, where all variables are strongly typed and there is no macro system.

It is perfectly reasonable to use private global constants, particularly for constanty things like URLs, default values, etc.

> I think this is fine since it's better to test from a clean slate each time anyway.

Constants are, by nature, immutable. Therefore each test is executing with the exact same data. They could have been instance properties, but then the properties would be re-instantiated for each test method. That's why I made them private global constants in the first place.

I would understand enabling this rule if it was a convention we already followed, but we don't. And I don't think we want to adopt it as a convention. Using it as a means of banning global constants altogether seems like an abuse of the rule.

qsmmorgan:
@philium Okay, up to you. I reverted the rule but it'd still be nice to replace a few of the global constants.

arcgis-runtime-samples-macos/Analysis/Line of sight (geoelement)/LineOfSightGeoElementViewController.swift
```diff
@@ -33,10 +30,12 @@ class LineOfSightGeoElementViewController: NSViewController {
     private let taxiGraphic: AGSGraphic
     private let observerGraphic: AGSGraphic
     private let lineOfSight: AGSGeoElementLineOfSight
+    
+    private let observerZMin = 20.0
```
philium: It might be nice to make this a range:
```swift
private let observerZRange = 20.0...1500.0
```

PR #262, opened 2018-11-14
Enables the identifier_name SwiftLint rule

qsmmorgan:


.swiftlint.yml
```diff
@@ -13,7 +13,6 @@ disabled_rules: # rule identifiers to exclude from running
   - for_where
   - force_cast
   - function_body_length
-  - identifier_name
```
philium: I'm going to have to think about this one for a little bit.

qsmmorgan: I was skeptical at first but I think it makes for better code.

- It flags underscored identifiers, some of which were still in the app
- It can make for more descriptive identifiers, like `segueID` for `id` or `layerIndex` for `i`
- The length constraints are configurable and exceptions can be made if needed

We can always disable it again in the future if it proves to be a pain.

philium: I like all of that. The only thing I'm concerned about is not knowing what the exceptions are without going to the config file. Occasionally there are members of the runtime team that help write samples. Allowing some short variable names and not others seems potentially confusing for them. Does the warning say what the exceptions are?

qsmmorgan: See https://github.com/Esri/arcgis-runtime-samples-ios/pull/581#issuecomment-439414236

arcgis-runtime-samples-macos/Maps/Manage operational layers/ManageOperationalLayersVC.swift
```diff
@@ -47,13 +47,13 @@ class ManageOperationalLayersVC: NSViewController, NSTableViewDataSource, NSTabl
         
     }
     
-    func moveLayer(_ layer: AGSLayer, from: Int, to: Int) {
-        self.mapView.map?.operationalLayers.removeObject(at: from)
+    func moveLayer(_ layer: AGSLayer, fromIndex: Int, toIndex: Int) {
```
philium: Not a big deal, but the new parameter labels should be `from` and `to`:

```swift
func moveLayer(_ layer: AGSLayer, from fromIndex: Int, to toIndex: Int) { … }
```

qsmmorgan: Done.

PR #269, opened 2018-11-21
Moves the Cut Geometry and Style WMS Layers controls 

qsmmorgan:
From the top right to bottom center.

Also changes the button in Symbolize Shapefile to the Push style.


philium:
On macOS it is most common for controls to be toward the top of a view rather than at the bottom. Why were these controls moved to below the map view?

PR #272, opened 2018-11-26
Adds a custom SwiftLint rule enforcing a single space between { and [ in a closure

qsmmorgan:


.swiftlint.yml
```diff
@@ -7,6 +7,10 @@ custom_rules:
     name: WGS84 Spatial Reference
     regex: AGSSpatialReference\(wkid:\s*(4326)\)
     message: "Use AGSSpatialReference.wgs84() instead."
+  closure_opening_space:
+    name: Closure Opening Space
+    regex: \{\[|\{\s{2,}\[
```
philium: This is a bit longer, but it prevents having to duplicate the opening curly brace and opening square bracket. I double checked that it matches no space and multiple spaces. Take it or leave it.
```suggestion
    regex: \{(\s{0}|\s{2,})\[
```

qsmmorgan: I like this change but I didn't want to wait for re-approval. We could do it in a separate PR.

philium: Is approval needed for suggestions? It seems like a suggested change has already been approved. I've been using the new suggestion feature to make suggestions, but I don't know what it is like to accept them or how that works.

qsmmorgan: So far I've needed approval for every suggested change I merged, I think. There may be circumstances where you wouldn't? I'm not sure.

PR #278, opened 2018-12-04
Adds the Buffer sample

qsmmorgan:


arcgis-runtime-samples-macos/Geometry/Buffer/BufferViewController.swift
```diff
@@ -0,0 +1,117 @@
...
+        guard bufferRadiusInMiles > 0 else {
+            return
+        }
+        
+        /// The buffer radius converted to meters.
+        let bufferRadiusInMeters = Measurement(value: bufferRadiusInMiles, unit: UnitLength.miles).converted(to: .meters).value
+        
+        /// The geometry for the map point buffered by the given distance in respect
+        /// to the projected map spatial reference system.
+        if let planarGeometry = AGSGeometryEngine.bufferGeometry(mapPoint, byDistance: bufferRadiusInMeters) {
```
philium: Were you trying to document the `planarGeometry` variable? Unfortunately the documentation comments aren't applied to the variable, so I would suggest just making them standard comments (`//`) here and before the following `if` statement.

screen shot 2018-12-05 at 11 03 07 am

arcgis-runtime-samples-macos/Geometry/Buffer/BufferViewController.swift
```diff
@@ -0,0 +1,117 @@
...
+        /// The user-set buffer radius in miles.
+        let bufferRadiusInMiles = textField.doubleValue
+        
+        // ensure that the buffer radius is a positive value
+        guard bufferRadiusInMiles > 0 else {
+            return
+        }
+        
+        /// The buffer radius converted to meters.
+        let bufferRadiusInMeters = Measurement(value: bufferRadiusInMiles, unit: UnitLength.miles).converted(to: .meters).value
```
philium: I like that you used the `Measurement` API to do the conversion. 👍

arcgis-runtime-samples-macos/Geometry/Buffer/BufferViewController.swift
```diff
@@ -0,0 +1,117 @@
...
+        
+        /// The geometry for the map point buffered by the given distance in respect
+        /// to the projected map spatial reference system.
+        if let planarGeometry = AGSGeometryEngine.bufferGeometry(mapPoint, byDistance: bufferRadiusInMeters) {
+            let graphic = AGSGraphic(geometry: planarGeometry, symbol: nil, attributes: nil)
```
philium: The `attributes` parameter has a default value of `nil`, so I would suggest omitting it wherever you create a graphic.

arcgis-runtime-samples-macos/Geometry/Buffer/README.md
```diff
@@ -0,0 +1,26 @@
...
+1. Click the map.    
+2. Graphics for a planar buffer (brown) and geodesic buffer (green) are added for the click location using the distance specified in the text box.
+3. Continue tapping to create additional buffers. Notice that buffers closer to the equator appear similar in size. As you move north or south from the equator, however, the geodesic polygons become much larger. Geodesic polygons are in fact a better representation of the true shape and size of the buffer.   
+4. Click `Clear Buffers` to remove all the graphics and start again.
+
+## How it works
+The map point for a tap on the display is captured and the buffer distance entered (in miles) is converted to meters. The static method `AGSGeometryEngine.bufferGeometry` is called to create a planar buffer polygon from the map location and distance. Another static method, `AGSGeometryEngine.geodeticBufferGeometry` is called to create a geodesic buffer polygon using the same inputs. The polygon results (and tap location) are displayed in the map view with different symbols in order to highlight the difference between the buffer techniques due to the spatial reference used in the planar calculation.
```
philium: Please use the full symbol names here. Omitting the parenthesis make the APIs look like properties.

```suggestion
The map point for a tap on the display is captured and the buffer distance entered (in miles) is converted to meters. The `AGSGeometryEngine` static method `bufferGeometry(_:byDistance:)` is called to create a planar buffer polygon from the map location and distance. Another static method, `geodeticBufferGeometry(_:distance:distanceUnit:maxDeviation:curveType:)` is called to create a geodesic buffer polygon using the same inputs. The polygon results (and tap location) are displayed in the map view with different symbols in order to highlight the difference between the buffer techniques due to the spatial reference used in the planar calculation.
```

The symbol names are longer (much longer in the second case), but it is the correct way of referring to Swift APIs.

arcgis-runtime-samples-macos/Geometry/Buffer/README.md
```diff
@@ -0,0 +1,26 @@
...
+
+## Relevant API
+* `AGSGeometryEngine.bufferGeometry(_:byDistance:)`
```
philium: Fully qualified symbol names for class methods include the term `class` to differentiate them from instance methods:

```suggestion
* `class AGSGeometryEngine.bufferGeometry(_:byDistance:)`
```

arcgis-runtime-samples-macos/Geometry/Buffer/README.md
```diff
@@ -0,0 +1,26 @@
...
+* `AGSGeometryEngine.geodeticBufferGeometry(_:distance:distanceUnit:maxDeviation:curveType:)`
```
philium:
```suggestion
* `class AGSGeometryEngine.geodeticBufferGeometry(_:distance:distanceUnit:maxDeviation:curveType:)`
```

arcgis-runtime-samples-macos/Geometry/Buffer/Buffer.storyboard
```diff
@@ -0,0 +1,98 @@
...
+                                                </textFieldCell>
+                                            </textField>
+                                            <textField verticalHuggingPriority="750" translatesAutoresizingMaskIntoConstraints="NO" id="GvW-du-Meg">
+                                                <rect key="frame" x="90" y="0.0" width="80" height="22"/>
+                                                <constraints>
+                                                    <constraint firstAttribute="width" constant="80" id="OdV-vs-CHR"/>
+                                                </constraints>
+                                                <textFieldCell key="cell" scrollable="YES" lineBreakMode="clipping" selectable="YES" editable="YES" sendsActionOnEndEditing="YES" borderStyle="bezel" title="300 mi" placeholderString="300 mi" drawsBackground="YES" id="JRL-CH-wwZ">
+                                                    <numberFormatter key="formatter" formatterBehavior="custom10_4" positiveFormat="# mi" usesGroupingSeparator="NO" lenient="YES" groupingSize="0" minimumIntegerDigits="0" maximumIntegerDigits="2000000000" id="yca-N5-9hu">
```
philium: Interesting choice to use a number formatter instead of a measurement formatter. Doing it this way lets you configure the formatting in Interface Builder, but it wouldn't work well in a localized environment. It also doesn't let you use the user's preferred format (based on their locale). It would be nice if we didn't have to use miles as the input unit for everyone.

arcgis-runtime-samples-macos/Geometry/Buffer/BufferViewController.swift
```diff
@@ -0,0 +1,117 @@
...
+        // instantiate a map using a topo basemap
+        let map = AGSMap(basemap: .topographic())
+        // assign the map to the map view
+        mapView.map = map
+        
+        // add the graphics overlays in the order we want them to draw, bottom to top
+        mapView.graphicsOverlays.addObjects(from: [geodesicOverlay, planarOverlay, clickLocationsOverlay])
+
+        // register to receive callbacks for pointer inputs in the map view
+        mapView.touchDelegate = self
```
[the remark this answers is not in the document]
qsmmorgan: I think in this case I'll leave it so developers can see it at a glance. A few other samples already show it being set in the storyboard.

arcgis-runtime-samples-macos/Geometry/Buffer/BufferViewController.swift
qsmmorgan: Thanks

arcgis-runtime-samples-macos/Geometry/Buffer/Buffer.storyboard
qsmmorgan: Good points. I've had trouble using measurement formatters in text fields on macOS before, though perhaps whatever issues I ran into have since been resolved.

qsmmorgan: Yeah, it looks like using a `MeasurementFormatter` for the `formatter` property of `NSTextField` doesn't work. I could still one and update things manually, but that seems like it would more of a distraction to developers than a useful feature. The actual unit of the buffer radius in the UI doesn't really matter much anyway. If you want me to make it work I will, but for now I'm marking this resolved.

PR #279, opened 2018-12-06
Adds the Update Attributes (Feature Service) sample

qsmmorgan:


arcgis-runtime-samples-macos/Edit data/Update attributes (feature service)/UpdateAttributesViewController.swift
```diff
@@ -0,0 +1,233 @@
...
+        super.init(coder: coder)
+    }
+
+    override func viewDidLoad() {
+        super.viewDidLoad()
+        
+        // create a map with a basemap
+        let map = AGSMap(basemap: .oceans())
+        // set initial viewpoint
+        map.initialViewpoint = AGSViewpoint(
```
philium: The initial viewpoint doesn't include any features. Was that intentional?

arcgis-runtime-samples-macos/Edit data/Update attributes (feature service)/UpdateAttributesViewController.swift
```diff
@@ -0,0 +1,233 @@
...
+    
+    private func update(feature: AGSFeature, damageType: String) {
+        
+        NSApp.showProgressIndicator()
+        
+        // update the damage type for the feature
+        feature.attributes[damageTypeAttributeKey] = damageType
+        
+        // update the values in the feature table with those we set in the feature attributes
+        featureTable.update(feature) { (error: Error?) in
```
philium: `self` should be captured weakly here.

arcgis-runtime-samples-macos/Edit data/Update attributes (feature service)/UpdateAttributesViewController.swift
```diff
@@ -0,0 +1,233 @@
...
+extension UpdateAttributesViewController: AGSCalloutDelegate {
+    
+    func didTap(_ callout: AGSCallout) {
+        
+        guard let featureForCallout = featureForCallout else {
+            return
+        }
+        
+        // hide the callout
+        callout.dismiss()
```
philium: Does the callout need to be dismissed here? The alert message is "Update the Selected Feature", but since the callout is dismissed, there is no selected feature. It isn't clear which feature is going to be updated.

arcgis-runtime-samples-macos/Content Display Logic/ContentPList.plist
```diff
@@ -834,6 +834,18 @@
 					<string>AddFeaturesViewController</string>
 				</array>
 			</dict>
+			<dict>
+				<key>displayName</key>
+				<string>Update attributes (feature service)</string>
+				<key>storyboardName</key>
+				<string>UpdateAttributesFeatureService</string>
+				<key>descriptionText</key>
+				<string>Tap on a feature to display a callout. The callout shows the damage type for that feature and an info button. To change the damage type tap on the info button and select a new on from the list shown</string>
```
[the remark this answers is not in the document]
qsmmorgan: Yep, this is from the iOS sample.

arcgis-runtime-samples-macos/Edit data/Update attributes (feature service)/UpdateAttributesViewController.swift
qsmmorgan: The initial viewpoint is from the iOS sample. Since this is a feature service where users can add and remove features, there are not always any in the initial viewpoint. I can still try to figure out a better solution.

arcgis-runtime-samples-macos/Edit data/Update attributes (feature service)/UpdateAttributesViewController.swift
qsmmorgan: Also from the iOS implementation, but good point.

arcgis-runtime-samples-macos/Edit data/Update attributes (feature service)/UpdateAttributesViewController.swift
```diff
@@ -0,0 +1,233 @@
...
+        mapView.map = map
+        
+        // set self for the delegates in order to receive callbacks
+        mapView.touchDelegate = self
+        mapView.callout.delegate = self
+    }
+    
+    private func damageType(for feature: AGSFeature) -> String {
+        // get the value from the attributes dictionary using the correct key
+        return feature.attributes[damageTypeAttributeKey] as! String
```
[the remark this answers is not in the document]
qsmmorgan: I'm fairly certain there will always be a value, but I made the function return an optional just in case. This way users can add add a value if one is missing.

arcgis-runtime-samples-macos/Edit data/Update attributes (feature service)/UpdateAttributesViewController.swift
qsmmorgan: I updated it to show most of the world, so features should be visible in most cases.

arcgis-runtime-samples-macos/Edit data/Update attributes (feature service)/UpdateAttributesViewController.swift
qsmmorgan: Seems like they were closing the callout during editing and then reopening it in order to update the title. I changed it so the callout title is updated directly.

qsmmorgan:
@philium Thanks for your suggestions. This should be ready for re-review.

PR #280, opened 2018-12-07
Adds missing app icon sizes

qsmmorgan:
Updates asset catalog to resuse the 512x512 image for both cases of it


qsmmorgan:
@philium Now that you mention it, I agree on having the same prefix. It's such minutia though that I'm not worried about it.

